PR #3, opened 2019-03-05
Open/Close file when necessary

NikhilNarayana:
Let me know if you think there is a better spot to put it, but since the user can stop needing the file at any point, I thought it might be better to just expose the option to close the fd on their end.


JLaferri:
Yeah this is definitely one option. I was thinking maybe instead just manually re-opening and closing the file whenever an operation runs that needs it instead.

PR #6, opened 2019-05-31
Add character colors and method to fetch them

mtimkovich:
Color information comes from here: https://supersmashbros.fandom.com/wiki/Palette_Swap_(SSBM)

### DIFF
--- a/src/melee/characters.js
+++ b/src/melee/characters.js
@@ -2,132 +2,132 @@ const externalCharacters = [{
   id: 0,
   name: "Captain Falcon",
   shortName: "Falcon",
-  colors: ["Black", "Red", "White", "Green", "Blue"],
+  colors: ["Default", "Black", "Red", "White", "Green", "Blue"],
 }, {
   id: 1,
   name: "Donkey Kong",
   shortName: "DK",
-  colors: ["Black", "Red", "Blue", "Green"],
+  colors: ["Default", "Black", "Red", "Blue", "Green"],
 }, {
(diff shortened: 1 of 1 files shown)


src/melee/characters.js
```diff
@@ -116,3 +142,17 @@ export function getCharacterShortName(externalCharacterId) {
   const character = getCharacterInfo(externalCharacterId) || {};
   return character.shortName;
 }
+
+// Return a human-readable color from a characterCode.
+export function getCharacterColorName(externalCharacterId, characterColor) {
+  const character = getCharacterInfo(externalCharacterId) || {};
+  if (!character || characterColor < 0 || characterColor > len(character)) {
```
JLaferri: what does the len function do and what is its purpose here? Should you be comparing it to `character.colors` instead?

JLaferri: A way to avoid doing this check might be to just use lodash's `get` function which will return undefined if the index is invalid

JLaferri: Actually I'd probably just add `"Default"` to all of the arrays as index zero and then write the function as:
```
const character = getCharacterInfo(externalCharacterId) || {};
const colors = character.colors || [];
return colors[characterColor] || null;
```

mtimkovich: > what does the len function do and what is its purpose here? Should you be comparing it to `character.colors` instead?

looks like I forgot what language I was writing in for a minute

PR #7, opened 2019-07-01
do getGameEnd in O(1) instead of O(n)

jmlee337:
we can count on the game end event being the last one in a completed replay

knowing if a game ended in time can help in filtering (IE. if a game timed-out in less than [however many minutes the ruleset says] then it wasn't a tournament match. This is relevant in Japan, where it's extremely common to do a 1 minute handwarmer in a timed mode)


src/SlippiGame.js
```diff
@@ -162,9 +162,12 @@ export default class SlippiGame {
       return this.gameEnd;
     }
 
-    // Trigger getFrames because that is where the flag is set
-    this.getFrames();
-    return this.gameEnd || null;
+    const slpfile = openSlpFile(this.input);
+
+    this.gameEnd = getGameEnd(slpfile);
```
JLaferri: Since getGameEnd can now return `{}` which is truthy, if this function is ever called on an incomplete file, and then later called when the file is complete, it will always return `{}`. Might be best to return null to avoid this?

I may also write a test that simulates this because this is a commonly overlooked case.

src/utils/slpReader.js
```diff
@@ -481,3 +484,22 @@ export function getMetadata(slpFile: SlpFileType): MetadataType {
...
+  if (gameEndLength <= 1 || slpFile.metadataLength <= 0) {
+    // This will happen on a severed incomplete file
+    // $FlowFixMe
+    return {};
+  }
+
+  const buffer = new Uint8Array(gameEndLength);
+  const start = slpFile.metadataPosition - METADATA_OFFSET - gameEndLength;
+  readRef(slpFile.ref, buffer, 0, gameEndLength, start);
+  const view = new DataView(buffer.buffer);
```
JLaferri: I might prefer to reuse existing code here? Might be especially useful if the game end payload ever changes sizes.

Something like:
```
const command = buffer[0];
const payload = buffer.slice(1);
return parseMessage(command, payload);
```

src/SlippiGame.js
JLaferri: I just pushed a test that should fail until this is fixed.

jmlee337: I forgot {} is truthy lol

jmlee337:
I think I might've rebased in a way that makes the diffs weird. Can remake if it's hard to read now

src/utils/slpReader.js
```diff
@@ -459,25 +461,39 @@ function readBool(view: DataView, offset: number): boolean | null {
...
 }
+
+export function getGameEnd(slpFile: SlpFileType): GameEndType | null {
+  if (slpFile.metadataLength <= 0) {
+    // This will happen on a severed incomplete file
+    // $FlowFixMe
+    return null;
+  }
+
+  const gameEndLength = slpFile.messageSizes[0x39] + 1;
```
JLaferri: 0x39 => Commands.GAME_END

src/utils/slpReader.js
```diff
@@ -459,25 +461,39 @@ function readBool(view: DataView, offset: number): boolean | null {
...
+  const buffer = new Uint8Array(gameEndLength);
+  const start = slpFile.metadataPosition - METADATA_OFFSET - gameEndLength;
+  readRef(slpFile.ref, buffer, 0, gameEndLength, start);
+  return parseMessage(buffer[0], buffer);
```
JLaferri: If buffer[0] is not 0x39 should we even call this function? Maybe just return null in that case? I'm slightly worried about garbage data having a command that exists in the first slot and the thing being processed weird.

jmlee337:
migrating to https://github.com/project-slippi/slp-parser-js/pull/8

PR #10, opened 2019-08-31
Migrate from Flow to Typescript

vinceau:
There is a clear trend towards Typescript for JS libraries over Flow static types. This PR migrates the project to use Typescript instead. Most types have been supported but there are still a few `any` types which needs to be squashed.

Merging this PR should close #9 since the Typescript definitions are built automatically.

This PR uses `rollup` to transpile from Typescript to Javascript and also updates the `README` with an npm badge and additional information for developers.


src/stats/actions.ts
```diff
@@ -1,59 +1,61 @@
...
+import { SlippiGame } from '../SlippiGame';
 
-function isRolling(animation) {
-  const rollAnimations = {
-    [States.ROLL_BACKWARD]: true,
-    [States.ROLL_FORWARD]: true,
-  };
-
-  return rollAnimations[animation];
+function isRolling(animation: State): boolean {
```
vinceau: The original code lacked clear typings, since it's supposed to return a boolean for but actually returns either `true` or undefined`.

```
function isRolling(animation) {
const rollAnimations = {
    [States.ROLL_BACKWARD]: true,
    [States.ROLL_FORWARD]: true,
};
return rollAnimations[animation];
}
```
Since you get the benefit of `enum` types in Typescript, it's clearer for the function type signature to specifically be `isRolling(animation): boolean`

JLaferri: I see, in that case the old code could be used but instead it could read:
```
  const rollAnimations = {
    [States.ROLL_BACKWARD]: true,
    [States.ROLL_FORWARD]: true,
  };

  return rollAnimations[animation] || false;
```

src/stats/overall.ts
```diff
@@ -112,10 +115,10 @@ function getBeneficialTradeRatio(conversionsByPlayerByOpening, playerIndex, oppo
   zippedTrades.forEach((conversionPair) => {
     const playerConversion = _.first(conversionPair);
     const opponentConversion = _.last(conversionPair);
-    const playerDamage = playerConversion.currentPercent - playerConversion.startPercent;
-    const opponentDamage = opponentConversion.currentPercent - opponentConversion.startPercent;
+    const playerDamage = (playerConversion as any).currentPercent - (playerConversion as any).startPercent;
+    const opponentDamage = (opponentConversion as any).currentPercent - (opponentConversion as any).startPercent;
 
-    if (playerConversion.didKill && !opponentConversion.didKill) {
+    if ((playerConversion as any).didKill && !(opponentConversion as any).didKill) {
```
JLaferri: What is the purpose of as any here? Does it not understand the types of the objects? There should be types with the currentPercent and didKill fields right?

src/stats/actions.ts
vinceau: Yes but I think that's much less readable than a simple switch-case statement. Not to mention that you need to instantiate a new object only to return one of the fields, and then to have to `|| false` in order to cast it to a boolean.

JLaferri: Very little in this case, I don't mind if you leave it. It's just a pattern I like cause I actually do find it more readable in more complex cases.

JLaferri: I also view optimizing for not instantiating objects and avoiding casts as micro-optimizations that are sort of irrelevant, compared to the stuff that actually slows down execution, it's worth the improved readability in my mind. I personally find it pretty readable, to me it reads like:

1. These are the roll animations
2. Is the current animation a roll animation? If not, false

But like I said, for the simple case both are pretty equivalent to me even from a readability standpoint.

vinceau: I think it would make more sense to store it as a list then makes more sense as opposed to an object. Like:
```
const rollAnimations = [States.ROLL_BACKWARD, States.ROLL_FORWARD];
return rollAnimations.contains(animation)
```

src/stats/overall.ts
nachoverdon: It's because the argument `conversionsByPlayerByOpening` is declared as `any`. Can be solved by declaring it on both `getOpeningRatio()` and `getBeneficialTradeRatio()` as ` { [x: string]: _.Dictionary<ConversionType[]> }`

```ts
type ConversionDictionary = { [x: string]: _.Dictionary<ConversionType[]> };
// ...
function getOpeningRatio(conversionsByPlayerByOpening: ConversionDictionary, playerIndex: number, opponentIndex: number, type: any): Ratio {}
// ...
function getBeneficialTradeRatio(conversionsByPlayerByOpening: ConversionDictionary, playerIndex: number, opponentIndex: number): Ratio {}
// ...
const playerDamage = playerConversion.currentPercent - playerConversion.startPercent;
const opponentDamage = opponentConversion.currentPercent - opponentConversion.startPercent;

if (playerConversion.didKill && !opponentConversion.didKill) {
// ...
```

vinceau: Yeah ideally there are no `any`s in the Typescript code. I've fixed this particular instance of any, but there are some `any`s that remain in the codebase. This PR actually doesn't really change much of the fundamental code or the logic, but fixing the remaining `any`s would require a reassessment of the actual typings and logic behind them.

vinceau: I'm not a fan of defining the `ConversionDictionary` type because that's a result of one of `lodash`'s magical functions and is not really a type. It would be better to cast it as an `any` for now and then reassess how to use proper typings such as `Map<string, ConversionType>` to ensure strict typing in the future.

src/stats/actions.ts
JLaferri: Once again, not a big deal with a list with 2 objects, but contains is an O(n) lookup. Anyway, we can keep the switch statement.

README.md
```diff
@@ -1,13 +1,26 @@
...
+npm install slp-parser-js
+```
+
+### Writing a simple script
+
 1) Create a fresh directory on your disk
-1) Inside this new directory, create a file called `script.js`
-1) Fill the `script.js` file with the following contents:
+2) Inside this new directory, create a file called `script.js`
+3) Fill the `script.js` file with the following contents:
```
JLaferri: This isn't necessary right? Github already auto-increments list numbers, makes it easier to inject new items in between existing list items.

src/SlippiGame.ts
```diff
@@ -143,13 +140,13 @@ export default class SlippiGame {
     return settings;
   }
 
-  getLatestFrame(): FrameEntryType | null {
+  getLatestFrame(): FrameEntryType {
```
JLaferri: Why does this no longer support null? It is possible that if this function is called before any data has been received in real-time mode that this function should return null

src/melee/moves.ts
```diff
@@ -152,16 +158,20 @@ const moves = {
   },
 };
 
-export function getMoveInfo(moveId) {
-  return moves[moveId];
+export function getMoveInfo(moveId: number): Move {
+  const m = moves[moveId];
+  if (!m) {
+    throw new Error(`Could not find move with id: ${moveId}`);
```
JLaferri: I don't like this. Our move definitions are not complete, being able to return "unknown" for undefined moves is actually used and this change would be a breaking change for some programs.

src/stats/common.ts
```diff
@@ -224,13 +219,13 @@ export function iterateFramesInOrder(
...
+  opponentIndices.forEach(indices => {
     initialize(indices);
 
     // Iterates through all of the frames for the current player and opponent
-    _.forEach(sortedFrames, (frame) => {
-      const playerPostFrame = _.get(frame, ['players', indices.playerIndex, 'post']);
-      const oppPostFrame = _.get(frame, ['players', indices.opponentIndex, 'post']);
+    sortedFrames.forEach(frame => {
+      const playerPostFrame = frame.players[indices.playerIndex].post;
+      const oppPostFrame = frame.players[indices.opponentIndex].post;
```
JLaferri: I think there was a reason this was previously written with get. Potentially in real-time, you might have received frame data from one player but not the other. I think this might cause an error in the new code because `players[indices.opponentIndex]` would return undefined.

src/stats/overall.ts
```diff
@@ -97,7 +101,8 @@ function getOpeningRatio(conversionsByPlayerByOpening, playerIndex, opponentInde
   return getRatio(openings.length, openings.length + opponentOpenings.length);
 }
 
-function getBeneficialTradeRatio(conversionsByPlayerByOpening, playerIndex, opponentIndex) {
+// FIXME: Stop using any here. The type is actually one of those wacky custom objects generated by lodash
```
JLaferri: I don't fully understand what this means. All lodash does is return standard js objects, what do you mean by wacky?

README.md
vinceau: That makes the assumption that users will only see the README on Github and all other programs which can open READMEs will also auto-increment lists. People could be viewing the README locally as a text file or something that's not Github. I personally think being correct (i.e. it is a list of steps which should increment at each step) outweighs the "ease of injecting new items" but that's just me. Will revert if you want.

src/stats/overall.ts
vinceau: Typescript is a superset of Javascript so it can't magically provide runtime type checking. It is only provides static type checking. Consider the following code:
```
const a = {
foo: 1,
bar: false,
baz: "hello world"
};
let str: string;
const x = _.get(a, [str]);
```
What type is x? How is your IDE meant to know what type `x` is so that it can warn you if you're accessing invalid object properties? It can't. You're right, Lodash does provide you a valid JS object, but as far as your IDE is concerned it has no idea what the type of `x` is which is what the `any` type is in Typescript. We can forgo all static type checking in code by saying they are of type `any` but it would best to avoid that since that defeats the purpose of type checking.

src/melee/moves.ts
vinceau: I've updated the `moves.ts` file with an `UnknownMove` constant.

src/SlippiGame.ts
vinceau: I've reverted back to the original code.

src/stats/common.ts
vinceau: I've reverted back to the original code.

src/stats/overall.ts
JLaferri: I see, that makes sense, that said I don't practically think we do anything quite like that in the codebase. In the case below I think you're referring to the `conversionsByPlayerByOpening` correct? I imagine the type of that is actually something like:
```
{
  [playerIndex: number]: {
    [openingType: string]: ConversionType[]
  }
}
```

JLaferri: Also with regards to a previous discussion, couldn't lodash types be specified on the variable they are initializing? In this case, for example:
```
const playerTrades: ConversionType[] = _.get(
  conversionsByPlayerByOpening, [playerIndex, 'trade']
 ) || [];
```

Though I do understand that it would be annoying to manually have to specify the type every time lodash it used.

src/stats/overall.ts
```diff
@@ -174,7 +187,8 @@ function generateInputs(game: SlippiGame) {
 
     // Do the same for c-stick
     const prevCstickRegion = getJoystickRegion(prevPlayerFrame.cstickX, prevPlayerFrame.cstickY);
-    const currentCstickRegion = getJoystickRegion(playerFrame.cstickX, playerFrame.cstickY);
+    // FIXME: stop using any. it is not clear what type thins is. cstickX does not exist on the type PreFrameUpdateType
+    const currentCstickRegion = getJoystickRegion((playerFrame as any).cstickX, (playerFrame as any).cstickY);
```
JLaferri: This actually looks like a bug lol. `cstickX` and `cstickY` do not exist on the pre-frame type, but `cStickX` and `cStickY` do.

JLaferri: And that right there would be one argument for using typescript lol

vinceau: Yeah I didn't want to change it to `cStickX` and `cStickY` since I wasn't sure if it would have any unintended consequences. Also since it's not really a Typescript issue the fix should probably not be a part of this PR.

src/stats/overall.ts
vinceau: As mentioned in [this thread](https://github.com/project-slippi/slp-parser-js/pull/10#discussion_r319748853) you can get around the `any` by specifying a `_.Dictionary` type (one of Lodash's typings for the returned value of `_.get()`) but you'd have to do that every time `_.get()` is used. For now, I think it's better to just cast it as an `any`.

nachoverdon:
🎉

PR #12, opened 2019-09-19
[PR] Calculate Stats on-the-fly

vinceau:
This PR adds calculation of statistics to be computed on a per-frame basis. This allows real-time stats calculation to be way more efficient.

The stat calculation involves implementing a `StatComputer` interface which maintains its own state and exposes a `processFrame()` method. This completely does away with needing to iterate over sorted frames.


src/stats/overall.ts
```diff
@@ -1,35 +1,22 @@
...
+import { ConversionType, PlayerIndexedType, StockType, OverallType, RatioType } from "./common";
+import { PlayerInput } from './inputs';
+
+export function generateOverallStats(
+  playerIndices: PlayerIndexedType[],
+  inputs: PlayerInput[],
+  stocks: StockType[],
+  conversions: ConversionType[],
+  playableFrameCount: number,
+): OverallType[] {
```
JLaferri: Worth mentioning that this function still increase in terms of processing times as the game gets longer. It wont be as bad as before but might be worth figuring out at some point how to make `overallStats` not reprocess everything every-time as well.

JLaferri:
So it seems to me like your method computes stats starting from the _process function. You're effectively hooking into the logic that makes iterateEvents not re-compute pre-computed data in order to make your stats stuff also not re-compute what has already been computed. Couple issues:

1) For files that are already complete, it no longer seems possible to call getSettings and have the lib only read the game settings section and nothing else. Even if all you need are settings, the lib will process the entire file regardless. This is highly problematic for the desktop app which can load thousands of replays at once to populate a table of games. This was... somewhat fast before because it only processed the game start payload from each file. Now it will parse everything and will likely be much much slower
2) Stats are now always computed even if you don't want them. Sometimes I might only want frame data and not subject myself to the performance loss of computing stats, this might be the case if I'm making a HUD replacement software.

src/utils/slpParser.ts
```diff
@@ -0,0 +1,114 @@
...
+        }
+
+        const location = command === Command.PRE_FRAME_UPDATE ? "pre" : "post";
+        const frames = payload.isFollower ? this.followerFrames : this.playerFrames;
+        this.latestFrameIndex = payload.frame;
+        _.set(frames, [payload.frame, 'players', payload.playerIndex, location], payload);
+        _.set(frames, [payload.frame, 'frame'], payload.frame);
+
+        const frame = frames[payload.frame];
+        this.statsComputer.processFrame(frame);
```
JLaferri: My understanding is that your method computes stats starting from the _process function. You're effectively hooking into the logic that makes iterateEvents not re-compute pre-computed data in order to make your stats stuff also not re-compute what has already been computed. Couple issues:

For files that are already complete, it no longer seems possible to call getSettings and have the lib only read the game settings section and nothing else. Even if all you need are settings, the lib will process the entire file regardless. This is highly problematic for the desktop app which can load thousands of replays at once to populate a table of games. This was... somewhat fast before because it only processed the game start payload from each file. Now it will parse everything and will likely be much much slower
Stats are now always computed even if you don't want them. Sometimes I might only want frame data and not subject myself to the performance loss of computing stats, this might be the case if I'm making a HUD replacement software.

src/stats/conversions.ts
```diff
@@ -1,221 +1,216 @@
...
-        }
-
-        if (state.move) {
-          state.move.hitCount += 1;
-          state.move.damage += opntDamageTaken;
+      this.state.set(indices, playerState);
+    })
+  }
+
+  public processFrame(frame: FrameEntryType, allFrames: FramesType): void {
```
JLaferri: Something I liked about my method that I had been working on is that in my equivalent to the "calculator" class, I had abstracted out the concept of the alternating player indices completely. Think it made for simpler-to-implement interface for the type that is most extensible in the system.

JLaferri: Perhaps this could be abstracted out via class inheritance somehow?

src/utils/slpParser.ts
JLaferri: Here are my (rough) thoughts on the two:
1. It might be possible to solve this by adding a simple arg to `_process` to allow it to short circuit on a getSettings call?
2. This should still be possible to tie to `getStats` but you'd have to manage the state for what the last frame that has been processed through the stats engine was externally somewhere. Basically every time `getStats` is called it iterates through all the frames that are new since the last `getStats` call?

src/utils/slpParser.ts
```diff
@@ -0,0 +1,114 @@
...
+    private followerFrames: FramesType | null = null;
+    private gameEnd: GameEndType | null = null;
+    private latestFrameIndex: number | null = null;
+    private statsComputer: Stats | null = null;
+    private playerIndices: PlayerIndexedType[] = [];
+
+    public constructor() {
+        this.playerFrames = {};
+        this.followerFrames = {};
+        this.statsComputer = new Stats([]);
```
JLaferri: Is it worth creating a Stats object here? Is this even a valid state for a Stats object? Might be better to not even initialize? Not sure.

vinceau: Yeah it's needed because you could call `getStats()` before you process any frames. The expected return value is the default stats object. But you can't even call `getStats()` if the object isn't initialized.

src/utils/slpParser.ts
vinceau: Yep I've added lazy processing of frames which is on by default. But I wanted the option to process stats on the fly as each frame is added so there's an option to enable that.

src/stats/stats.ts
```diff
@@ -0,0 +1,85 @@
...
+    public process(): void {
+        // Finish processing if we're not up to date
+        if (this.lastFrame === this.lastProcessedFrame) {
+            return;
+        }
+        if (this.opponentIndices.length === 0) {
+            return;
+        }
+        let i = this.lastProcessedFrame ? this.lastProcessedFrame + 1 : Frames.FIRST;
+        while (Boolean(this.frames[i])) {
```
JLaferri: Probably not necessary to use `Boolean()` here, when you put something like this in a while loop it's already implied that it's being interpreted as a boolean. Not blocking or important, but just seems a bit too explicit

src/stats/stats.ts
```diff
@@ -0,0 +1,85 @@
...
+
+export class Stats {
+    private options: StatOptions;
+    private lastProcessedFrame: number | null = null;
+    private lastFrame: number;
+    private frames: FramesType = {};
+    private opponentIndices = new Array<PlayerIndexedType>();
+    private allComputers = new Array<StatComputer<unknown>>();
+
+    public constructor(options?: StatOptions) {
```
JLaferri: I don't really think this option is necessary. It's not settable by the user of the lib atm is it?

In any case, if we do want to keep it, usually I feel like what I'm used to seeing when libraries do this is that each field in the options object is optional. Basically something like:
```
export interface StatOptionsInput {
    processOnTheFly: boolean?;
}
```
And then using: `public constructor(options: StatOptionsInput = {})`
What do you think?

src/stats/stats.ts
```diff
@@ -0,0 +1,85 @@
...
+                return;
+            }
+            this.allComputers.forEach(comp => comp.processFrame(frame, this.frames));
+            this.lastProcessedFrame = i;
+            i++;
+        }
+    }
+
+    public addFrame(frame: FrameEntryType): void {
+        this.frames[frame.frame] = frame;
```
JLaferri: Feels weird to need this, we already have this in the location where `process` is called from SlippiGame. Why not just pass the frames in as an arg to `process` instead of re-managing the same object in this class?

src/stats/stats.ts
```diff
@@ -0,0 +1,85 @@
...
+        this.allComputers.push(computer);
+    }
+
+    public registerAll(computers: StatComputer<unknown>[]): void {
+        this.allComputers = this.allComputers.concat(computers);
+    }
+
+    public process(): void {
+        // Finish processing if we're not up to date
+        if (this.lastFrame === this.lastProcessedFrame) {
```
JLaferri: This feels mostly unnecessary. In this situation, the while loop below will always be false. I kind of worry that the management of these two might become slightly messed up in the future and this condition might cause bugs. Not blocking

src/stats/stocks.ts
```diff
@@ -1,26 +1,42 @@
...
+  stock: StockType | null | undefined;
+}
+
+export class StockComputer implements StatComputer<StockType[]> {
+  private state = new Map<PlayerIndexedType, StockState>();
+  private opponentIndices = new Array<PlayerIndexedType>();
+  private stocks = new Array<StockType>();
+
+  public setPlayerIndices(playerIndices: PlayerIndexedType[]): void {
+    this.opponentIndices = playerIndices;
```
JLaferri: The naming convention for these really confuses me in a lot of places. `playerIndices` gets set to `opponentIndices`? Truth is I struggled with naming these too. Really when it's in the array form its like `playerPermutations` or something, then when it's the single object it's `indices` maybe?

src/SlippiGame.ts
```diff
@@ -72,185 +48,128 @@ export class SlippiGame {
...
         return false;
       }
 
       switch (command) {
-      case Command.GAME_START:
-        payload = payload as GameStartType;
-        if (!payload.stageId) {
-          return true; // Why do I have to do this? Still not sold on Flow
-        }
+        case Command.GAME_START:
```
JLaferri: I personally prefer the previous style of indenting. It kind of bothers me that with this form, switch statements double-indent logic whereas an `if/else` statement would have only single indented the logic. Do you have a preference? If so why?

src/stats/stats.ts
vinceau: Removed the `Boolean` assertion in commit 5016f1.

src/stats/stats.ts
vinceau: Removed the options in commit ee6424.

src/stats/stats.ts
vinceau: Not sure what this comment is referring to. But I definitely don't like the idea of passing in frames into process because that relies on the caller to have access to the frames. Where as with stats having its own state of the frames it knows where the processing is up to and it can just internally process the newer frames if needed.

src/SlippiGame.ts
vinceau: Nope. I used the built-in auto formatter. I'm sure there's some style guide or some standard that the formatter is using so I just used it to clean up the code. No particular reason but imo using the auto-formatter is easy so I use it.

src/stats/stats.ts
vinceau: Removed in commit 2f41eb.

src/SlippiGame.ts
JLaferri: Yeah, agreed. Do you know if the auto-formatter uses the eslint rules or something?

src/stats/stocks.ts
vinceau: Yeah you're telling me. I'm just as confused. The function `getSinglesOpponentIndices()` returns a type called `PlayerIndexedType`. I don't know whether it should be called player indices or opponent indices. In any case I've just named them all `opponentIndices` for now. Fixed in commit 85d076.

src/SlippiGame.ts
vinceau: No idea. Probably some vscode setting or something though. Haven't looked into it.

src/stats/stocks.ts
vinceau: No particular preference. Preferable if names are something short and succinct.

src/utils/slpParser.ts
```diff
@@ -1,108 +1,104 @@
...
+
+      switch (payload.internalCharacterId) {
+      case 0x7:
+        playersByIndex[playerIndex].characterId = 0x13; // Sheik
+        break;
+      case 0x13:
+        playersByIndex[playerIndex].characterId = 0x12; // Zelda
+        break;
+      }
+      this.settingsComplete = true;
```
JLaferri: This re-introduces a bug I had fixed when I changed how the short circuiting worked. Probably important to understand.

Settings are not complete (pre 1.6.0) until all of the post frame updates have come through for the first frame, not just the first post frame update. This is used to set the character properly in the case of a sheik transform. The way the code is written now, if sheik played in the higher port number, the settings would should them as Zelda.

`this.settingsComplete = payload.frame > Frames.FIRST` could be placed outside the if statement and it would work properly. This causes it to wait for the first frame update of frame -122, meaning that all -123 post frames have come through.

vinceau: Fixed in commit b801d6.

PR #16, opened 2019-10-21
Fix Typing Issues

jkosh44:


src/stats/common.ts
```diff
@@ -92,6 +92,12 @@ export type OverallType = PlayerIndexedType & {
   beneficialTradeRatio: RatioType;
 };
 
+export type ConversionsByPlayerByOpening = {
+  [playerIndex: string]: {
+    [openingType: string]: ConversionType[]
+  }
+}
```
JLaferri: I feel like this doesn't need to be exported as a common type, it can probably just be declared locally in the file where it's used?

jkosh44: Good point, I've moved it to the top of `overall.ts`.

PR #23, opened 2020-04-23
Add interactiveOpponent to combo objects

NikhilNarayana:
provide a boolean that lets the user know if there were no inputs from the opponent at all.

Goal is to provide a way for users to know if they should filter combo because there was no interaction from the combo'd player (possibly because they were away from their controller)

This only covers the 0% change in inputs case. There is some discussion to be had about values between 0 and 100% that might be worth notifying the user about or if we should provide something else entirely.

### DIFF
--- a/src/stats/combos.ts
+++ b/src/stats/combos.ts
@@ -1,9 +1,9 @@
 import _ from 'lodash';
-import { PostFrameUpdateType } from "../utils/slpReader";
+import { PostFrameUpdateType, PreFrameUpdateType } from "../utils/slpReader";
 import { FrameEntryType, FramesType, MoveLandedType, ComboType, PlayerIndexedType } from "./common";
 import {
   isDamaged, isGrabbed, calcDamageTaken, isTeching, didLoseStock,
-  Timers, isDown, isDead
+  Timers, isDown, isDead, isInputting
 } from "./common";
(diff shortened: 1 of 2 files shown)


vinceau:
In the current iteration, if a player holds left 100 frames before a combo to DI or whatever and does not change inputs for the duration of the combo, the `interactiveOpponent` field would still be false.

If instead, you were to track the last input of a player, you would detect a change in the X/Y value of the control stick, and set the last input frame to be `comboStartFrame-100`. Then people can later decide whether to include a combo or not because the opponent was still pressing something "recently" for whatever definition of recent.

NikhilNarayana:
> In the current iteration, if a player holds left 100 frames before a combo to DI or whatever and does not change inputs for the duration of the combo, the `interactiveOpponent` field would still be false.

Vince and I already discussed this a bit and I said the chance of this happening is so low in 45%+ combos so i don't see it as a big issue.

vinceau:
> I do like this idea in general, but it requires users to pick a threshold to accept and I don't think we have the data to help users pick.

Naively you could just check that the last frame where the opponent pressed something is greater than zero. This would still allow filtering of the games/combos where the player is not present (assuming the user did not leave mid-game), while reducing the chance of filtering out a genuine combo to be close to zero.

PR #25, opened 2020-06-26
Add WinnerIndex to OverallType

a-sink-a-wait:
Mainly looking for feedback; am I on the right track here? Don't want to commit any more time if I am way off base.

I've implemented a `winnerIndex` property from my understanding of just digging through the code. This is in an effort to display the winner in the replays folder of the desktop application.

### DIFF
--- a/src/stats/overall.ts
+++ b/src/stats/overall.ts
@@ -32,7 +32,7 @@ export function generateOverallStats(
     const conversions = _.get(conversionsByPlayer, playerIndex) || [];
     const successfulConversions = conversions.filter(conversion => conversion.moves.length > 1);
     const opponentStocks = _.get(stocksByPlayer, opponentIndex) || [];
-    const playerStocks = _.get(stocksByPlayer, playerIndex);
+    const playerStocks = _.get(stocksByPlayer, playerIndex) || [];
     const opponentEndedStocks = _.filter(opponentStocks, 'endFrame');
 
     const conversionCount = conversions.length;
@@ -48,6 +48,7 @@ export function generateOverallStats(
       conversionCount: conversionCount,
       totalDamage: totalDamage,
       killCount: killCount,
+      winnerIndex: determineWinner(playerStocks, opponentStocks),
 
       successfulConversions: getRatio(successfulConversionCount, conversionCount),
       inputsPerMinute: getRatio(inputCount, gameMinutes),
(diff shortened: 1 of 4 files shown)


src/stats/overall.ts
```diff
@@ -67,6 +69,23 @@ export function generateOverallStats(
   return overall;
 }
 
+function determineWinner(playerStocks: StockType[], opponentStocks: StockType[]) : number {
+
+  if (playerStocks.length == 0 && opponentStocks.length == 0) return -1; //unknown winner;
```
jwsharpe, line 74: can we avoid returning undefined by taking out the if statement and moving the `return -1` to the end of the function, will avoid weird edge cases or missing data errors later.

a-sink-a-wait: Sounds good!

a-sink-a-wait:
Updated this. Still need tests and some cleanup. Different approach, need some feeback.

vinceau:
For starters, I'm not really sure I agree with having the winner index as a part of the overall stats object. Since it's not really a stat. 

But some comments on the code itself: I think it would be better to export a helper function that gets imported instead of having the player determination code in the `SlippiGame` class directly. Additionally you could probably use a reduce instead of a `forEach` [see here for inspiration](https://github.com/vinceau/slp-realtime/blob/master/src/utils/winner.ts).

The key issue with adding the winner (I assume for the purpose of displaying in the desktop app) is that it requires processing the entire file for all the frames `getStats()` which is not performant. People have already complained about the desktop app being slow already and at the moment, it only loads the metadata. This is the primary reason why winner is not included at the moment.

a-sink-a-wait:
@vinceau Solid feedback. A `reduce` is definitely possible. I wanted something working and reviewable.

Is this worth implementing at the moment? Should we hold off?

vinceau:
@a-sink-a-wait I think it would be best to hold off for the time being. Fizzi has talked about adding caching of sorts and pagination which would allow processing entire SLP files less costly. But until then, even if this was to be merged you probably wouldn't see it's use in the desktop app because the performance hit would be too costly.

a-sink-a-wait:
I know there's this PR as well that aims to improve performance of the desktop App.
Would it be worth trying to combine these two implementations?

NikhilNarayana:
> I know there's this PR as well that aims to improve performance of the desktop App. [project-slippi/slippi-desktop-app#58](https://github.com/project-slippi/slippi-desktop-app/pull/58)
> 
> Would it be worth trying to combine these two implementations?

It would be worth know what kind of performance we would be getting if combined.

a-sink-a-wait:
I'll take a look into this.

NikhilNarayana:
closing due to placements being in game end now

PR #30, opened 2020-07-22
save buffers to array for relay

NikhilNarayana:
Save an array of buffers containing new data and then only concat when getFullBuffer is called.


src/utils/slpStream.ts
```diff
@@ -80,6 +84,9 @@ export class SlpStream extends Writable {
     // Join the current data with the old data
     const data = Uint8Array.from(Buffer.concat([this.previousBuffer, newData]));
 
+    // Save the new data into the bufferArray so we can join it if necessary
+    this.buffersToConcat.push(newData);
```
JLaferri, line 88: This should probably get cleared on the start of a new game right?

NikhilNarayana:
Actually, I'm just gonna implement it in the desktop app completely.

PR #36, opened 2020-10-10
fix offsets and add extra fields to item update

NikhilNarayana:


vinceau:
The expiration timer type should be a float not an int.

What do you think about calling it `chargePower` instead of `chargedPower`? I feel it sounds better and also matches with `chargeShotLaunched`.

vinceau (approved):

PR #40, opened 2020-10-24
Added analog metrics.

WattTheFlux:
Added in tracking the analog distance for the joystick in addition to computing the distance traveled in meters and the average velocity of the stick throughout the game. Analog distance is measured in joystick coordinate units while meters and meters per second are used for the real world distances. After looking at the average velocity, it's not as interesting as a metric as I initially thought it'd be. It's pretty low just due to the fact that in general, more frames are spent holding a direction than moving it. Maybe it'd be more interesting to track joystick acceleration or to exclude frames that don't have any movement? Eh.

### DIFF
--- a/src/stats/common.ts
+++ b/src/stats/common.ts
@@ -83,6 +83,9 @@ export interface OverallType extends PlayerIndexedType {
   successfulConversions: RatioType;
   inputsPerMinute: RatioType;
   digitalInputsPerMinute: RatioType;
+  joystickDistanceTraveled: number;
+  joystickMetersTraveled: number;
+  joystickAverageVelocity: RatioType;
   openingsPerKill: RatioType;
   damagePerOpening: RatioType;
   neutralWinRatio: RatioType;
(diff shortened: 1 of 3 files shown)


src/stats/common.ts
```diff
@@ -83,6 +83,9 @@ export interface OverallType extends PlayerIndexedType {
   successfulConversions: RatioType;
   inputsPerMinute: RatioType;
   digitalInputsPerMinute: RatioType;
+  joystickDistanceTraveled: number;
```
NikhilNarayana, line 86: Instead of having these all at the top level, make a joystick section and create a new JoystickType, similar to the InputCountsType in usage. Doesn't have to be called JoystickType, I just don't have a better suggestion atm.

NikhilNarayana (changes requested):
Looks good functionality wise, obviously the one question mark is size of the stick gate. If we could reference some authority, that would help me feel more confident.

NikhilNarayana:
closing due to age

PR #41, opened 2020-11-06
update PostFrame with everything except stage bits

NikhilNarayana:
Updated PostFrameUpdateType to match the spec (mostly). Might want to update the tests.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -72,7 +72,21 @@ export interface PostFrameUpdateType {
   lastHitBy: number | null;
   stocksRemaining: number | null;
   actionStateCounter: number | null;
+  miscActionState: number | null;
+  isAirborne: boolean | null;
+  lastGroundID: number | null;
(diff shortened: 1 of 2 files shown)


src/types.ts
```diff
@@ -72,7 +72,21 @@ export interface PostFrameUpdateType {
   lastHitBy: number | null;
   stocksRemaining: number | null;
   actionStateCounter: number | null;
+  miscActionState: number | null;
+  isAirborne: boolean | null;
+  lastGroundID: number | null;
```
vinceau, line 77:
```suggestion
  lastGroundId: number | null;
```
The rest of the SDK uses lowercase `d` in `Id` so we should probably change it to that to be consistent.

vinceau (changes requested):

vinceau (approved):

vinceau:
LGTM

PR #46, opened 2021-01-07
Fix bad lint for fs.writeSync invocation in slpFile.ts

jmlee337:
whoever wrote the line originally mixed up the two method signatures for writeSync. The reason it worked to begin with is that the length argument is ignore if not a number.

add a test to ensure correctness, since non-trivial effort is required to test manually.


vinceau (approved):
LGTM. Thanks for this!

NikhilNarayana (approved):

PR #47, opened 2021-01-09
Add missing fields to MetadataType interface

jmlee337:
MetadataType.players[].names appears in replay files from netplay and is documented in the [replay spec](https://github.com/project-slippi/slippi-wiki/blob/master/SPEC.md#players-metadata), but does not appear in the MetadataType interface.

Add the field to the interface and add a test to read it

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -133,6 +133,7 @@ export interface MetadataType {
       };
     };
   } | null;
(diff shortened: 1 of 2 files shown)


jmlee337:
I also found that consoleNick was missing from MetadataType

NikhilNarayana (approved):

PR #53, opened 2021-01-31
Read display name and connect code from game start if present

jmlee337:
in the case of file writing, write display name and connect code to metadata if present in game start

contingent on project-slippi/Ishiiruka#231 and project-slippi/slippi-ssbm-asm#51 which will enact the updates to replay spec


jmlee337:
prerequisite PRs are merged, this is ready for review

jmlee337:
friendly bump!

src/types.ts
```diff
@@ -18,6 +18,8 @@ export interface PlayerType {
   teamId: number | null;
   controllerFix: string | null;
   nametag: string | null;
+  displayName: string;
+  connectCode: string;
```
vinceau, line 22: Should these not be nullable fields too? Especially from replays created on console or in VS Mode?

src/utils/slpFile.ts
```diff
@@ -1,16 +1,26 @@
...
+  players: {
+    [playerIndex: number]: {
+      characterUsage: {
+        [internalCharacterId: number]: number;
+      };
+      names: {
+        netplay: string;
+        code: string;
+      };
+    };
```
vinceau, line 22: Again, shouldn't these values or the entire `names` object be nullable?

test/filewriter.spec.ts
```diff
@@ -1,41 +1,88 @@
...
-      const timeoutMs = 1000;
+      setTimeout(() => {
+        const players = new SlippiGame(newFilename).getMetadata().players;
+        Object.keys(players).forEach((key) => {
+          const player = players[key];
+          players[key] = player.names;
+        });
+        fs.unlinkSync(newFilename);
+
+        expect(players).toEqual({ 0: { netplay: "", code: "" }, 1: { netplay: "", code: "" } });
```
vinceau, line 36: I'm not sure if this is good practice or personal preference, but I think if we know something does not exist, it should be null, rather than an empty string. Like `player.names` should return null for replays from console/vs mode. Thoughts on having `null` as opposed to the `{ netplay: string, code: string }` object?

vinceau (changes requested):
The code itself looks fine overall, but would like some discourse surrounding nullable names/codes to happen first.

test/filewriter.spec.ts
jmlee337: my understanding of best practices is that we should avoid `null` and nullable fields/vars wherever possible. I assumed that turning on strict null checks in https://github.com/project-slippi/slippi-js/commit/f3c4ca997c518c8924f45fe2176c0ac622471ceb expressed a preference toward removing `null` from the codebase completely in the long run

vinceau: That's sound enough logic for keeping them not nullable.

A different concern I had was there currently exists a `player.nametag` field which contains their in-game name tag if applicable (in offline/VS mode). Should these fields not be combined together? Whether that looks like or something else:
```
names: {
  netplay: "",
  tag: "",
  code: "",
}
```
I'm still not entirely sold on any of these names maybe aside from "code". Since in-game it's called `Name entry` so maybe the in-game name should be called `name`, but then having `player.names.name` field seems kinda really dumb.

jmlee337: you mean in `types.ts`?

what purpose would that serve?

src/utils/slpFile.ts
```diff
@@ -1,16 +1,26 @@
...
   lastFrame: number;
-  players: any;
+  players: {
+    [playerIndex: number]: {
+      characterUsage: {
+        [internalCharacterId: number]: number;
+      };
+      names: {
+        netplay: string;
+        code: string;
```
vinceau, line 20:
```suggestion
      netplay: {
        name: string;
        code: string;
```

src/utils/slpFile.ts
```diff
@@ -101,6 +111,22 @@ export class SlpFile extends Writable {
...
+          if (player.type === 3) {
+            return;
+          }
+
+          this.metadata.players[i] = {
+            characterUsage: {},
+            names: {
+              netplay: player.displayName,
+              code: player.connectCode,
+            },
```
vinceau, line 126:
```suggestion
            netplay: {
              name: player.displayName,
              code: player.connectCode,
            },
```

src/utils/slpFile.ts
```diff
@@ -221,7 +247,33 @@ export class SlpFile extends Writable {
...
+      footer = Buffer.concat([footer, Buffer.from("U"), Buffer.from([5]), Buffer.from("names{")]);
+
+      // Write display name
+      footer = Buffer.concat([
+        footer,
+        Buffer.from("U"),
+        Buffer.from([7]),
+        Buffer.from("netplaySU"),
+        Buffer.from([player.names.netplay.length]),
+        Buffer.from(`${player.names.netplay}`),
```
vinceau, line 263: Can you make the top level: `netplay` with the attributes `netplay.name` and `netplay.code`?

test/filewriter.spec.ts
vinceau: I think my gripe was with `names` being the top level key but then the in-game tag is also a name but isn't in the object, and technically `code` isn't really a name. But if we just rename the top level key to `netplay` and the in netplay display name field to `name` I'd be fine with it.

test/filewriter.spec.ts
```diff
@@ -1,41 +1,88 @@
...
+          players[key] = player.names;
+        });
         fs.unlinkSync(newFilename);
 
-        expect(writtenDataLength).toBe(dataLength);
+        expect(players).toEqual({
+          0: { netplay: "ekans", code: "EKNS#442" },
+          1: { netplay: "gaR's uncle", code: "BAP#666" },
+          2: { netplay: "jmlee337", code: "JMLE#166" },
+          3: { netplay: "Mr.SuiSui", code: "SUI#244" },
```
vinceau, line 58: Again please make these `netplay.name` and `netplay.code`.

vinceau (changes requested):

test/filewriter.spec.ts
jmlee337: these are already existing fields in MetadataType. Changing MetadataType fields would break compatibility with existing replay files.

In this light, do you still want me to make the corresponding change in slpFile.ts? That type is internal to the file and can be changed.

vinceau (approved):

test/filewriter.spec.ts
vinceau: Sadge

PR #54, opened 2021-02-02
Doubles Stats

OGoodness:
Added the ability to calculate doubles stats. 
Treat every other player on the stage as an opponent during combo.

Change perspective of combos/conversions to be from the victims point of view. 
Combos now belong to the victim, you can identify whos combo it is by checking the moveIds. 
Conversions now also belong to the victim.

Can calculate stats for free-for-all, teams, and 1v1. 
Overall, the character handling should be more generalized. 

Also added test cases for `src/melee` and `src/console`


src/stats/overall.ts
```diff
@@ -58,9 +67,9 @@ export function generateOverallStats(
       digitalInputsPerMinute: getRatio(inputCounts.buttons, gameMinutes),
       openingsPerKill: getRatio(conversionCount, killCount),
       damagePerOpening: getRatio(totalDamage, conversionCount),
-      neutralWinRatio: getOpeningRatio(conversionsByPlayerByOpening, playerIndex, opponentIndex, "neutral-win"),
-      counterHitRatio: getOpeningRatio(conversionsByPlayerByOpening, playerIndex, opponentIndex, "counter-attack"),
-      beneficialTradeRatio: getBeneficialTradeRatio(conversionsByPlayerByOpening, playerIndex, opponentIndex),
+      neutralWinRatio: response.map((item) => item.neutralWinRatio),
+      counterHitRatio: response.map((item) => item.counterHitRatio),
+      beneficialTradeRatio: response.map((item) => item.beneficialTradeRatio),
```
vinceau, line 72: I'm not too sure about these being an array for each player's ratio against each opponent. What do you think about aggregating these into a single ratio per player, instead of for each head-to-head? Like say it's a free for all game, shouldn't their neutral win ratio simply be averaged across their other opponents? Or say it's a team game, they should probably be averaged across their non-team mates?

Also I've removed almost all mentions of opponent indices. The overall stats is now the only place left where these still exists such a concept of opponent indices. It might be good if we can remove even that reference.

vinceau (commented):
For the most part I'm very happy with this PR. Only have some thoughts and comments regarding the overall stats.

src/stats/overall.ts
OGoodness: Thanks for that! I saw those changes, sorry that you had to make those.

vinceau (approved):
LGTM!

PR #57, opened 2021-02-15
Fix edge-case stats calculation

OGoodness:
This PR fixes two critical issues with stats calculation:
1. command grabs are not included in conversions
2. total damage calculation is too naive and does not account for damage recovery, pichu's self damage, and blast zone damage

New Slippi files are included for tests including:
- Bowser Side Special and variations
- Falcon Hug
- Ganon Hug
- Kirby Swallow/Release/Breakout
- Yoshi Egg Lay


test/stats.spec.ts
```diff
@@ -13,4 +13,76 @@ describe("when calculating stats", () => {
     expect(p2Success).toBe(5);
     expect(p2Fail).toBe(4);
   });
+  describe("When Handling Command Grabs", () => {
+    it("accounts for Bowser's Command Grab", () => {
+      const game = new SlippiGame("slp/consistencyTest/BowsVDK-SB-63.slp");
+      const stats = game.getStats();
+      const bowser = stats.overall[0]
+      let totalDamageBowserDealt = 0
+      stats.conversions.map( conversion => {
```
vinceau, line 22:
```suggestion
      stats.conversions.forEach( conversion => {
```
Minor pet-peeve, can you replace all the `map` with `forEach`? In these cases we don't really care about the actual mapping of one array into another array, rather we just care about iterating through each value.

vinceau (changes requested):

test/stats.spec.ts
vinceau: You'd need to repeat this change for all the other `.map` instances btw. Not just this one instance.

src/stats/common.ts
```diff
@@ -189,13 +201,23 @@ export function isDown(state: number): boolean {
...
   return state >= State.CAPTURE_START && state <= State.CAPTURE_END;
 }
 
+// TODO: Find better implementation of 3 seperate ranges
+export function isCommandGrabbed(state: number): boolean {
+  return (
+    ((state >= State.COMMAND_GRAB_RANGE1_START && state <= State.COMMAND_GRAB_RANGE1_END) ||
+      (state >= State.COMMAND_GRAB_RANGE2_START && state <= State.COMMAND_GRAB_RANGE2_END) ||
+      (state >= State.COMMAND_GRAB_RANGE3_START && state <= State.COMMAND_GRAB_RANGE3_END)) &&
+    state !== State.BARREL_CANNON_WAIT
```
NikhilNarayana, line 217: why not just have `COMMAND_GRAB_RANGE2_END` go up to `0x152`?

src/stats/common.ts
```diff
@@ -137,6 +137,18 @@ export enum State {
   FALL_BACKWARD = 0x1f,
   GRAB = 0xd4,
   CLIFF_CATCH = 0xfc,
+  DAMAGE_FALL = 0x26,
+
+  // Command Grabs
+  BARREL_CANNON_WAIT = 0x153,
```
vinceau, line 143: Did you mean to explicitly ignore the BarrelWait or BarrelCannonWait?

```suggestion
  BARREL_WAIT = 0x125,
```

src/stats/common.ts
```diff
@@ -137,6 +137,18 @@ export enum State {
...
+  COMMAND_GRAB_RANGE1_START = 0x10a,
+  COMMAND_GRAB_RANGE1_END = 0x130,
+
+  COMMAND_GRAB_RANGE2_START = 0x147,
+  COMMAND_GRAB_RANGE2_END = 0x153,
```
vinceau, line 148: As Nikki mentioned already, why set the end range to `0x153` and then explicitly ignore `0x153`? Why not just set the end range to `0x152`?

```suggestion
  COMMAND_GRAB_RANGE2_END = 0x152,
```

vinceau (changes requested):

vinceau (approved):

PR #73, opened 2021-05-26
adding wall tech/ground tech action counts

Ckoenig1:
just added two new objects to the action counts to track wall techs and tech directions along with the code to update them.


test/stats.spec.ts
```diff
@@ -7,6 +7,12 @@ const expectedThrow = {
   back: 2,
   down: 1,
 };
+const expectedTech = {
+  backward: 2,
+  forward: 1,
+  neutral: 8,
+  fail: 4,
+};
```
vinceau, line 15: Can you move this inside the tech test function itself like in line 145?

vinceau (changes requested):

test/stats.spec.ts
Ckoenig1: yep

vinceau (approved):

PR #77, opened 2021-06-23
Allow SlippiGame to accept an ArrayBuffer

frankborden:
This makes it easier to use slippi-js in the browser via the Snowpack build tool, and probably for other build tools as well.

Snowpack can polyfill dependencies that need Node globals like Buffer by inlining them into the module of the dependency. That means slippi-js gets it's own Buffer library for internal use, but it's not possible for app code to create a Buffer object that will satisfy the instanceof check even if they use the same polyfill library. Passing in an ArrayBuffer will work on Node and in the browser.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@slippi/slippi-js",
-  "version": "6.1.0",
+  "version": "6.1.1",
   "description": "Official Project Slippi Javascript SDK",
   "license": "LGPL-3.0-or-later",
   "repository": "project-slippi/slippi-js",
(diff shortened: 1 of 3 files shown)


JLaferri:
With `create-react-app` slippi-js works fine out of the box using Node Buffers in the browser. Is there some reason your app can't do the same?

frankborden:
As mentioned in the discord, `create-react-app` does a polyfill to let you use Buffer in your app code because it's not node. With other web build tools this can be more difficult or not possible, even if they will polyfill the slippi-js library itself. I think this change expands support a good amount without compromising any of the internals or doing the extra work to rewrite all buffer usage to be natively compatible with browsers.

vinceau (approved):

PR #82, opened 2021-10-13
feat: add getRollbackFrames()

NikhilNarayana:
Tested against altf4's rollback counter and the results match up.


src/utils/slpParser.ts
```diff
@@ -266,3 +282,42 @@ export class SlpParser extends EventEmitter {
     }
   }
 }
+
+class RollbackCounter {
```
vinceau, line 286: Can we move this into a separate class? Perhaps `src/utils/RollbackCounter.ts` or something?

src/utils/slpParser.ts
```diff
@@ -266,3 +282,42 @@ export class SlpParser extends EventEmitter {
...
+      this.lastFrameWasRollback = true;
+    } else if (this.lastFrameWasRollback) {
+      this.rollbackLengths.push(this.currentRollbackLength);
+      this.currentRollbackLength = 0;
+      this.lastFrameWasRollback = false;
+    }
+    return this.lastFrameWasRollback;
+  }
+
+  public get(): RollbackFrames {
```
vinceau, line 320: I'm not sure i like the naming of `.get()`. `rollbackCounter.get()` sounds like to me it would return just the count. I think if we had like `.getFrames()`, `.getCount()`, and `.getLengths()` as 3 separate methods, it would be better, that way we don't care if the shape of `RollbackFrames` changes.

src/utils/slpParser.ts
```diff
@@ -122,6 +126,10 @@ export class SlpParser extends EventEmitter {
     return this.frames;
   }
 
+  public getRollbackFrames(): RollbackFrames {
+    return this.rollbackCounter.get();
```
vinceau, line 130: This would need to be updated to make the 3 different calls to `rollbackCounter` to fit the shape of `RollbackFrames`.

vinceau (approved):
LGTM

PR #86, opened 2021-12-15
Add grab and various item moves to moves list

cnkeats:
Also adds a test suite for various move ids


NikhilNarayana:
closing due to age

PR #91, opened 2022-01-22
feat/expand-types

cnkeats:
This PR covers a few different improvements to some types and their descriptions.

- Expands `character.ts` exports to include missing descriptions for the characters that were added in #88 
- Expands the `GameMode` type to include most in-game scenes.
  -  Home Run Contest and Target Test scenes will start showing up in replays after https://github.com/project-slippi/slippi-ssbm-asm/pull/87 is merged
  - The rest of the scenes were added for future-proofing, since they could eventually end up being recorded in replays too
  - Includes a test to verify that the scene is being detected properly.
-  Adds more stages to the stage list, also for future-proofing

Attached are replays for all 51 events that were used for verification for their stages. Many of the replays will not play correctly due to various mechanics currently unsupported by playback, but the stages will still be valid.

[event-1-20.zip](https://github.com/project-slippi/slippi-js/files/7917189/event-1-20.zip)
[event-21-30.zip](https://github.com/project-slippi/slippi-js/files/7917186/event-21-30.zip)
[event-31-40.zip](https://github.com/project-slippi/slippi-js/files/7917187/event-31-40.zip)
[event-41-51.zip](https://github.com/project-slippi/slippi-js/files/7917188/event-41-51.zip)

PR #93, opened 2022-01-29
Add stadium support

cnkeats:
This PR expands many aspects with the goal of adding support to the Slippi Launcher for the Target Test and Home-Run Contest game modes.

- Adds `frameToGameTimer()` in `utils` for converting a frame number to the in-game timer value for that frame
- Greatly expands the [game info block](https://github.com/project-slippi/slippi-wiki/blob/master/SPEC.md#game-info-block) in `getSettings()`
- Adds more ids for various 1-player stages
  - Replays for event stages can be found in #91 
- Adds more ids for various game modes with a test for verification


src/types.ts
```diff
@@ -24,17 +31,68 @@ export interface PlayerType {
...
   players: PlayerType[];
   scene: number | null;
   gameMode: GameMode | null;
+  randomSeed: number | null;
+  isFrozenPS: boolean | null;
+  minorScene: number | null;
+  majorScene: number | null;
+}
+
+export interface GameInfo {
```
NikhilNarayana, line 74:
```suggestion
export interface GameInfoType {
```
for consistency

src/utils/gameTimer.ts
```diff
@@ -0,0 +1,44 @@
+import { GameStartType, TimerType } from "..";
+
+export function frameToGameTimer(frame: number, settings: GameStartType): string {
+  const timerType = ((settings.gameInfo?.gameBitfield1 as number) & 0b11) as TimerType;
+
+  if (timerType == TimerType.DECREASING) {
+    let remainingFrames = frame > 0 ? frame : 0;
+    const startingTimerInFrames = (settings.gameInfo?.gameTimer as number) * 60;
+
+    const minutesPassed = Math.floor(remainingFrames / 60 / 60);
```
NikhilNarayana, line 10: I'm not set on this idea but you could create a couple anonymous functions for calculating minutes, seconds, and centiseconds based on frames since the logic is the same for increasing or decreasing.

src/utils/gameTimer.ts
```diff
@@ -0,0 +1,44 @@
+import { GameStartType, TimerType } from "..";
+
+export function frameToGameTimer(frame: number, settings: GameStartType): string {
```
NikhilNarayana, line 3: I think this should take in GameInfoType instead of GameStartType

NikhilNarayana (changes requested):

src/utils/gameTimer.ts
cnkeats: I'm not a fan of this since it would break them into multiple functions and would end up taking more space. For centiseconds, the calculation is different based on the timer type.

vinceau: 1. Why is this function so complicated? Could this logic not simply be a matter of computing the duration by converting into seconds, and then formatting that using something like `moment.format("mm:ss:SS")`
2. If this function must be computed manually instead of using a library, I think would prefer two seperate functions `frameToCountdownTimer(frame: number, startingDurationFrames: number)`, `frameToCountupTimer(frame: number)`.

test/timer.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+    expect(frameToGameTimer(timer_limit.getMetadata()?.lastFrame as number, timer_limit_gameinfo)).toBe("00:00.00");
+  });
+
+  it("should calculate the timer correctly when decreasing", () => {
+    const timer_decreasing = new SlippiGame("./slp/timer_decreasing.slp");
+    const timer_decreasing_gameinfo = timer_decreasing.getSettings()?.gameInfo as GameInfoType;
+    expect(frameToGameTimer(timer_decreasing.getMetadata()?.lastFrame as number, timer_decreasing_gameinfo)).toBe(
+      "01:51.76",
+    );
+  });
```
vinceau, line 25: You don't need to parse an entire`SLP` file just to test the timer logic. You can simply pass a `lastFrame` number and duration to the conversion function.

vinceau (changes requested):

src/types.ts
```diff
@@ -24,17 +31,68 @@ export interface PlayerType {
...
+  minorScene: number | null;
+  majorScene: number | null;
+}
+
+export interface GameInfoType {
+  gameBitfield1: number | null;
+  gameBitfield2: number | null;
+  gameBitfield3: number | null;
+  gameBitfield4: number | null;
+  bombRain: boolean | null;
```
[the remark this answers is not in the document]
cnkeats: yes!

src/utils/slpReader.ts
```diff
@@ -305,26 +305,63 @@ export function parseMessage(command: Command, payload: Uint8Array): EventPayloa
...
+      const getEnabledItems = (): number => {
+        const offsets = [0x1, 0x100, 0x10000, 0x1000000, 0x100000000];
+        const enabledItems = offsets.reduce((acc, offset) => {
+          const byte = readUint8(view, 0x29) as number;
+          return acc + byte * offset;
+        }, 0);
+
+        return enabledItems;
+      };
+
```
vinceau, line 343: Maybe make this helper function accept `view` as a paramenter and then move it outside of the current function scope.

test/timer.spec.ts
```diff
@@ -0,0 +1,22 @@
+import { GameStartType, SlippiGame } from "../src";
+import { frameToGameTimer } from "../src/utils/gameTimer";
+
+describe("when calculating the in-game timer", () => {
+  it("should calculate the timer correctly when increasing", () => {
+    const timer_increasing = new SlippiGame("./slp/timer_increasing.slp");
+    const timer_increasing_gameinfo = timer_increasing.getSettings();
+    expect(frameToGameTimer(2014, timer_increasing_gameinfo as GameStartType)).toBe("00:33.57");
```
vinceau, line 8: This test file is really only testing the `frameToGameTimer` conversion function. Because of that, you don't need to even declare `new SlippiGame()` or access `game.getSettings()`. Just pass in a mock `GameStartType` and delete the test SLPs.

You can also probably just co-locate this test with the `gameTimer.ts` file in the same `utils` folder but called `gameTimer.spec.ts`.

src/types.ts
```diff
@@ -12,30 +12,140 @@ export interface PlayerType {
...
   connectCode: string;
   userId: string;
 }
 
-export enum GameMode {
+export enum GameModeType {
   VS = 0x02,
+  CLASSIC = 0x03,
+  ADVENTURE = 0x04,
+  ALL_STAR = 0x05,
```
vinceau, line 44: @JLaferri, I remember asking you about this before, about whether or not the contants exported in `slippi-js` should be exhaustive and represent a general Melee JS library or Slippi-specific JS library, and I recall you said something along the lines of Slippi and original Melee has divered in some ways so you'd prefer it to just be Slippi-specific constants e.g. 0x08 is not used normally, so we use it for online.

I don't know if you recall this conversation and whether or not you have thoughts about this PR expanding all these constants to include all the original Melee modes/constants.

src/utils/slpReader.ts
```diff
@@ -305,26 +305,63 @@ export function parseMessage(command: Command, payload: Uint8Array): EventPayloa
...
-          nametag,
-          displayName,
-          connectCode,
-          userId,
+          staminaMode: !!readUint8(view, 0x6c + playerIndex * 0x24, 0x01),
+          silentCharacter: !!readUint8(view, 0x6c + playerIndex * 0x24, 0x02),
+          lowGravity: !!readUint8(view, 0x6c + playerIndex * 0x24, 0x04),
+          invisible: !!readUint8(view, 0x6c + playerIndex * 0x24, 0x08),
+          blackStockIcon: !!readUint8(view, 0x6c + playerIndex * 0x24, 0x10),
+          metal: !!readUint8(view, 0x6c + playerIndex * 0x24, 0x20),
```
vinceau, line 319: I prefer wrapping things in `Boolean(value)` instead of `!!value` for boolean coersion since I think it's clearer.

src/types.ts
```diff
@@ -12,30 +12,140 @@ export interface PlayerType {
...
+export enum EnabledItemType {
+  METAL_BOX = 2 ** 0,
+  CLOAKING_DEVICE = 2 ** 1,
+  POKEBALL = 2 ** 2,
+  // Bits 4 through 8 of item bitfield 1 are unknown
+  UNKNOWN_ITEM_BIT_4 = 2 ** 3,
+  UNKNOWN_ITEM_BIT_5 = 2 ** 4,
+  UNKNOWN_ITEM_BIT_6 = 2 ** 5,
+  UNKNOWN_ITEM_BIT_7 = 2 ** 6,
+  UNKNOWN_ITEM_BIT_8 = 2 ** 7,
```
vinceau, line 112: How come you went with defining the values here as `2 ** X` instead of just writing out the number?

src/utils/slpReader.ts
```diff
@@ -305,26 +305,63 @@ export function parseMessage(command: Command, payload: Uint8Array): EventPayloa
...
+          rumbleEnabled: !!readUint8(view, 0x6c + playerIndex * 0x24, 0x80),
+          cpuLevel: readUint8(view, 0x74 + offset),
+          offenseRatio: readFloat(view, 0x7d + offset),
+          defenseRatio: readFloat(view, 0x81 + offset),
+          modelScale: readFloat(view, 0x85 + offset),
+          controllerFix: controllerFix,
+          nametag: nametag,
+          displayName: displayName,
+          connectCode: connectCode,
+          userId: userId,
```
vinceau, line 330: Use object shorthand notation here:

```suggestion
          controllerFix,
          nametag,
          displayName,
          connectCode,
          userId,
```

src/utils/gameTimer.ts
```diff
@@ -0,0 +1,21 @@
+import type { GameStartType } from "..";
+import { TimerType } from "..";
+import { format } from "date-fns";
+
+export function frameToGameTimer(frame: number, gameStart: GameStartType): string {
```
vinceau, line 5: I think you can simplify this function type signature. If we're only converting numbers to strings essentially and doing formatting, we probably don't need to know about the entirety of the `GameStartType`. Really it only needs to know whether or not the game timer is decreasing and if so, what the starting timer was.

So you could probably just do something like:

```suggestion
export function frameToGameTimer(frame: number, startingTimer?: number): string {
```
If `startingTimer` is passed in, then assume the timer is decreasing, otherwise assume that the timer is increasing.

vinceau (changes requested):
This PR looks okay to me for the most part. But I'm interested to know what @JLaferri thinks about exhaustively expanding all these constants.

src/utils/slpParser.ts
```diff
@@ -116,6 +118,24 @@ export class SlpParser extends EventEmitter {
...
+  public getItems(): EnabledItemsType | null {
+    if (this.settings?.itemSpawnBehavior == ItemSpawnBehaviorType.OFF) {
+      return null;
+    }
+
+    const itemBitfield = this.settings?.enabledItems;
+    const enabledItems = {} as EnabledItemsType;
+    enabledItems.items = [];
+
+    for (let i = 0; i < 40; i++) {
```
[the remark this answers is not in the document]
cnkeats: 5 bytes of item bits. I'll add a declaration.

src/types.ts
```diff
@@ -12,30 +12,140 @@ export interface PlayerType {
...
+  modelScale: number | null;
   controllerFix: string | null;
   nametag: string | null;
   displayName: string;
   connectCode: string;
   userId: string;
 }
 
-export enum GameMode {
+export enum GameModeType {
```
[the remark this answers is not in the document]
cnkeats: See @NikhilNarayana's review above - it was changed for consistency

src/types.ts
cnkeats: I did it this way for readability. I don't want to write out the numbers in decimal since they get pretty large and it may not be immediately obvious to someone digging through the library where the numbers come from - `2^39` is `549755813888`. Writing out as binary works but is also long - 40 characters for the above number.

src/utils/gameTimer.ts
cnkeats: I'm not a fan of this solution since we also have the option to have the timer disabled. I can add an extra parameter for timer type if you think that's a good compromise.

src/types.ts
cnkeats: As part of my exploration of stadium replays, I ended up recording replays for every game mode. A lot of them don't work properly, but there are plenty in other modes that already work; I figured I'd add the rest while I was here since it's possible I work on other modes down the line - it's fine to add them when we officially support those replays though.

JLaferri:
> This PR looks okay to me for the most part. But I'm interested to know what @JLaferri thinks about exhaustively expanding all these constants.

I don't mind it, I think it's probably better to have more info than less. That said, it looks like there's a few breaking changes? Field names have changed. Example: https://github.com/project-slippi/slippi-js/pull/93/files#diff-c54113cf61ec99691748a3890bfbeb00e10efb3f0a76f03a0fd9ec49072e410aR69

So we'd probably need to increase the major version number.

src/utils/gameTimer.ts
vinceau: By "timer disabled", is that different from `TimerType.NONE`?

This function still shouldn't depend on the entirety of the `GameStartType` object. So you can just pick the required fields instead.
```suggestion
export function frameToGameTimer(frame: number, options: Pick<GameStartType, "timerType" | "startingTimerSeconds">): string {
```

src/utils/gameTimer.ts
```diff
@@ -0,0 +1,21 @@
...
+  const timerType = gameStart.timerType;
+
+  if (timerType == TimerType.DECREASING) {
```
vinceau, line 8:
```suggestion
  if (timerType === TimerType.DECREASING) {
```

src/utils/gameTimer.ts
```diff
@@ -0,0 +1,21 @@
...
+    const centiseconds = Math.ceil((((60 - (frame % 60)) % 60) * 99) / 59);
+    const date = new Date(0, 0, 0, 0, 0, (gameStart.startingTimerSeconds as number) - frame / 60, centiseconds * 10);
+    return format(date, "mm:ss.SS");
+  }
+
+  if (timerType == TimerType.INCREASING) {
```
vinceau, line 14:
```suggestion
  if (timerType === TimerType.INCREASING) {
```

src/utils/gameTimer.spec.ts
```diff
@@ -0,0 +1,31 @@
+import { GameStartType, TimerType } from "..";
+import { frameToGameTimer } from "./gameTimer";
+
+describe("when calculating the in-game timer", () => {
+  it("should calculate the timer correctly when increasing", () => {
+    const increasing_settings = {
+      timerType: TimerType.INCREASING,
+      startingTimerSeconds: 0,
+    } as GameStartType;
```
vinceau, line 9:
```suggestion
    };
```
If we pick the required fields as in my other suggestion we don't need these casts either.

src/utils/gameTimer.ts
```diff
@@ -0,0 +1,21 @@
...
+    const centiseconds = Math.floor(((frame % 60) * 99) / 59);
+    const date = new Date(0, 0, 0, 0, 0, frame / 60, centiseconds * 10);
+    return format(date, "mm:ss.SS");
+  }
+
+  return "";
```
vinceau, line 20: Do you think we should probably return something else here instead. Maybe like "Infinite" or "Endless" or something?

src/types.ts
```diff
@@ -53,6 +81,86 @@ export interface FrameStartType {
...
+  BARREL = 2 ** 34,
+  EGG = 2 ** 35,
+  PARTY_BALL = 2 ** 36,
+  BARREL_CANNON = 2 ** 37,
+  BOMB_OMB = 2 ** 38,
+  MR_SATURN = 2 ** 39,
+}
+
+export interface EnabledItemsType {
+  items: string[] | null;
```
vinceau, line 161:
```suggestion
  items: string[] | null;
```
Why is this an array of strings? Also why do we even need this type? It seems kinda useless. Can't we just return `EnabledItemType[]`?

src/types.ts
```diff
@@ -53,6 +81,86 @@ export interface FrameStartType {
...
+export enum ItemSpawnBehaviorType {
+  OFF = 0xff,
+  VERY_LOW = 0x00,
+  LOW = 0x01,
+  MEDIUM = 0x02,
+  HIGH = 0x03,
+  VERY_HIGH = 0x04,
+}
+
+export enum EnabledItemType {
```
vinceau, line 116:
```suggestion
export enum ItemType {
```
Can we just call this `ItemType`?

src/types.ts
```diff
@@ -53,6 +81,86 @@ export interface FrameStartType {
...
+  damageRatio: number | null;
+}
+
+export enum TimerType {
+  NONE = 0b00,
+  DECREASING = 0b10,
+  INCREASING = 0b11,
+}
+
+export enum ItemSpawnBehaviorType {
```
vinceau, line 107:
```suggestion
export enum ItemSpawnType {
```
And can we shorten this one too?

src/utils/slpParser.ts
```diff
@@ -125,6 +130,24 @@ export class SlpParser extends EventEmitter {
     return this.settingsComplete ? this.settings : null;
   }
 
+  public getItems(): EnabledItemsType | null {
```
vinceau, line 133:
```suggestion
  public getItems(): EnabledItemType[] {
```

src/types.ts
```diff
@@ -15,20 +15,36 @@ export interface PlayerType {
...
+  modelScale: number | null;
   controllerFix: string | null;
   nametag: string | null;
   displayName: string;
   connectCode: string;
   userId: string;
 }
 
-export enum GameMode {
+export enum GameModeType {
```
vinceau, line 43:
```suggestion
export enum GameMode {
```
Can we revert this and not make breaking changes where possible? Ideally this doesn't have to be a breaking change.

src/types.ts
```diff
@@ -38,13 +54,25 @@ export enum Language {
...
+  selfDestructScoreValue: number | null;
   stageId: number | null;
+  startingTimerSeconds: number | null;
+  enabledItems: number | null;
+  damageRatio: number | null;
   players: PlayerType[];
-  scene: number | null;
-  gameMode: GameMode | null;
+  minorScene: number | null;
+  majorScene: number | null;
```
vinceau, line 70:
```suggestion
  scene: number | null;
  majorScene: number | null;
```
Likewise over here, can we just keep minor scene as scene?

src/types.ts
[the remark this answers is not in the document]
vinceau: You're adding a new type. When you added a new type, you should be consistent. Changing an existing type makes it a breaking change.

src/types.ts
```diff
@@ -38,13 +54,25 @@ export enum Language {
...
+  friendlyFireEnabled: boolean | null;
+  bombRainEnabled: boolean | null;
   isTeams: boolean | null;
-  isPAL: boolean | null;
+  itemSpawnBehavior: ItemSpawnBehaviorType | null;
+  selfDestructScoreValue: number | null;
   stageId: number | null;
+  startingTimerSeconds: number | null;
+  enabledItems: number | null;
+  damageRatio: number | null;
```
vinceau, line 67: Why are there so many duplicate fields? Whatever fields exist in `GameInfoType` can we remove from `GameStartType`?

vinceau (changes requested):
If you can remove the breaking changes, it will be easier for this to be merged.

src/utils/gameTimer.spec.ts
cnkeats: I haven't used Pick before - that's a nice feature and I agree that this works better

src/utils/gameTimer.ts
cnkeats: Originally I wrote this to get the exact display of the timer and added the empty string to for full coverage - if there isn't one then there is nothing. It could potentially cause confusion but I think that scenario would be really rare. Of the suggestions, I like "Infinite" but it really does not matter to me.

src/types.ts
cnkeats: There is a distinction to be made between game items and enabled items in the settings.

In my upcoming Stadium computers, the in-game item's type is used to pick out the targets from the game data. We could also use it to give items names based on their type, such as lasers and turnips.

I'm fine with changing this to `ItemType` if needed.

NikhilNarayana (approved):

vinceau (changes requested):
Thanks for fixing those breaking changes. This is looking much, much better. Just a few more small nits and then it should good to merge.

vinceau (approved):
LGTM! Thanks for your contribution :)

PR #94, opened 2022-01-31
feat: parse and add slippi UIDs to player info

NikhilNarayana:


src/types.ts
```diff
@@ -20,6 +20,7 @@ export interface PlayerType {
   nametag: string | null;
   displayName: string;
   connectCode: string;
+  slippiUid: string;
```
vinceau, line 23: Can we just call this `userId`?
```suggestion
  userId: string;
```

src/utils/slpReader.ts
```diff
@@ -303,6 +313,7 @@ export function parseMessage(command: Command, payload: Uint8Array): EventPayloa
           nametag: nametag,
           displayName: displayName,
           connectCode: connectCode,
+          slippiUid: slippiUid,
```
vinceau, line 316: Not your fault really, but we should just use object shorthand notation here.

vinceau (changes requested):

vinceau (approved):
LGTM

PR #95, opened 2022-02-01
feat: stadium calculations

cnkeats:
Adds target break and homerun distance computers


src/stats/homerun.ts
```diff
@@ -0,0 +1,43 @@
+import { EventEmitter } from "events";
+
+import type { FrameEntryType, FramesType } from "../types";
+import { positionToHomerunDistance } from "../utils/homerunDistance";
+import type { StatComputer } from "./stats";
+
+export class HomerunComputer extends EventEmitter implements StatComputer<number> {
+  private gameDistance = 0;
+
+  public constructor() {
```
NikhilNarayana, line 10: you can remove the constructor, the default one will call super for you.

NikhilNarayana (changes requested):

src/SlippiGame.ts
```diff
@@ -1,4 +1,5 @@
 import type { StatOptions, StatsType } from "./stats";
+import { HomerunComputer, TargetBreakComputer } from "./stats";
```
vinceau, line 2:
```suggestion
import { HomeRunComputer, TargetBreakComputer } from "./stats";
```
Can we capitalise the `Run` in `homeRun` and `HomeRun`?

src/SlippiGame.ts
```diff
@@ -181,6 +187,12 @@ export class SlippiGame {
       actionCounts: this.actionsComputer.fetch(),
       overall: overall,
       gameComplete,
+      targetBreaks: settings.gameMode === GameMode.TARGET_TEST ? this.targetBreakComputer.fetch() : null,
+      // homerun distance depends on the language setting (not NTSC/PAL) and JPN has not been implemented
+      homerunDistance:
+        settings.gameMode === GameMode.HOME_RUN_CONTEST && settings.language !== Language.JAPANESE
+          ? this.homerunComputer.fetch()
+          : null,
```
vinceau, line 195: I don't really think these belong in the `getStats()` call or the `StatsType` object. They're very unrelated to the rest of the stats which mainly only make sense in a competitive environment. I would prefer like a separate call `getStadiumStats()` which could return a `StadiumStats` type instead.

src/stats/homerun.ts
```diff
@@ -0,0 +1,39 @@
+import { EventEmitter } from "events";
+
+import type { FrameEntryType, FramesType } from "../types";
+import { positionToHomerunDistance } from "../utils/homerunDistance";
+import type { StatComputer } from "./stats";
+
+export class HomerunComputer extends EventEmitter implements StatComputer<number> {
```
vinceau, line 7: I don't see the benefit of this `StatComputer` since the distance would increase by a little bit every frame, and a new event fired every frame. Whereas the only information that we would care about is the final result. I think we should remove this stat computer entirely, and just calculate the distance on `getStadiumStats()` call based on the last frame we've received.

src/utils/homerunDistance.ts
```diff
@@ -0,0 +1,26 @@
+// Returns the recorded homerun distance in FEET given the in-game position of the sandbag.
+// The conversion from internal game units to feet and meters is not currently perfectly understood.
```
vinceau, line 2: When you say not perfectly understood, does that mean that we don't know the conversion logic at all? Is it not possible currently to make this function accept like a param for the unit to return it in?
e.g. `positionToHomeRunDistance(distance: number, units: 'feet' | meters' = 'feet')`?

vinceau (changes requested):

src/utils/homerunDistance.ts
cnkeats: The math is... not simple. The Stadium community doesn't have a perfectly accurate formula to convert from in-game units to displayed distance. They have a spreadsheet [here](https://docs.google.com/spreadsheets/d/11K1QI_sZi8ePWvejQAG-2uhvBEi7zTdBluscIFPl6Ow/edit#gid=0) that I was pointed to for the math. It works perfectly to most values but some large ones are off.

I've added a conversion from feet to meters.

cnkeats: taukhan has since been in contact with me and has supplied me with a much simpler formula for calculating score, which I have updated this file with

the reason their calculations were so complex is because they were keeping track of records in Google Sheets which does not support float32

src/stats/common.ts
```diff
@@ -11,6 +11,11 @@ export interface StatsType {
   overall: OverallType[];
 }
 
+export interface StadiumStatsType {
+  targetBreaks: TargetBreakType[] | null;
+  homeRunDistance: number | null;
+}
+
```
vinceau, line 18:
```suggestion
export type StadiumStatsType = HomeRunContestResultType | TargetTestResultType; 

export interface TargetTestResultType {
  type: 'target-test',
  targetBreaks: TargetBreakType[];
}

export interface HomeRunContestResultType {
  type: 'home-run-contest',
  distance: number;
  units: 'feet' | 'meters';
}

```
The `StadiumStatsType` should probably be discriminated union.

src/utils/homeRunDistance.ts
```diff
@@ -0,0 +1,23 @@
+export function positionToHomeRunDistance(distance: number, units: "feet" | "meters" = "feet"): number {
+  // In NTSC vs. PAL, meters are different sizes
+  const feetModeConversionFactor = 0.952462;
+  const metersModeConversionFactor = 1.04167;
```
vinceau, line 4:
```suggestion
  const FEET_CONVERSION_FACTOR = 0.952462;
  const METERS_CONVERSION_FACTOR = 1.04167;
```
Can we change these constants casing and move them out of the function and to the top of the file?

src/SlippiGame.ts
```diff
@@ -194,6 +207,43 @@ export class SlippiGame {
...
+      return null;
+    }
+
+    this.stadiumStatsComputer.process();
+
+    let sandbag = null;
+    for (let i = 0; i < settings.players.length; i++) {
+      sandbag = players[i]?.post.internalCharacterId === 32 ? players[i]?.post : null;
+    }
+
```
vinceau, line 231:
```suggestion
  switch (settings.gameMode) {
    case GameMode.TARGET_TEST:
      // Return TargetTestResultType
      break;
    case GameMode.HOME_RUN_CONTEST:
      // Return HomeRunContestResultType
      break;
    default:
      return null;
  }
```
Can we use a switch-case here?

vinceau (changes requested):

src/SlippiGame.ts
cnkeats: updated to use a switch statement - please let me know if there is a cleaner way to filter to the `Sandbag` character

src/SlippiGame.ts
```diff
@@ -69,13 +74,21 @@ export class SlippiGame {
       this.inputComputer,
       this.stockComputer,
     );
+
+    // Set up stadium stats calculation
+    this.stadiumStatsComputer = new Stats(opts);
+    this.stadiumStatsComputer.register(this.targetBreakComputer);
```
vinceau, line 80: Can we not have a second `stadiumStatsComputer`? I don't think we need it.
We can just register the `this.targetBreakComputer` with `this.statsComputer`.

src/stats/targets.ts
```diff
@@ -0,0 +1,76 @@
...
+  private targetBreaks = new Array<TargetBreakType>();
+
+  public constructor() {
+    super();
+  }
+
+  public setup(): void {
+    // Reset the state
+    this.targetBreaks = [];
+  }
```
vinceau, line 21: `setup()` gets a copy of the `GameSettingsType`. We should determine whether or not this current game is a target test game and if not do nothing on `processFrame()`.

E.g. we can do this:
```
  private isTargetTestGame: boolean = false;
```
And then in the `setup()` function:

```
this.isTargetTestGame = settings.gameMode === GameMode.TARGET_TEST;
```
Then in `processFrame()` we can just return early if it's not a target test game.

vinceau (changes requested):

src/index.ts
```diff
@@ -7,6 +7,7 @@ export * from "./types";
 
 // Utils
 export * from "./utils/gameTimer";
+export * from "./utils/homeRunDistance";
```
vinceau, line 10: I don't think we need to export this.

test/homeRun.spec.ts
```diff
@@ -0,0 +1,18 @@
...
+    const homeRunDistance = positionToHomeRunDistance(inGameUnits);
+    expect(homeRunDistance).toBe(0);
+  });
+
+  it("should correctly calculate the distance for positive-distance hits", () => {
+    const game = new SlippiGame("slp/homeRun_positive.slp");
+    // get sandbag's final X position
+    const inGameUnits = game.getLatestFrame()?.players[1]?.post.positionX as number;
+    const homeRunDistance = positionToHomeRunDistance(inGameUnits);
+    expect(homeRunDistance).toBeCloseTo(1070.9, 0);
```
vinceau, line 16: Can we test this via the call to `getStadiumStats` instead?

Can we also include a test that the units change based on the game language?

test/targets.spec.ts
```diff
@@ -0,0 +1,11 @@
+import { SlippiGame, TargetTestResultType } from "../src";
+
+describe("when processing break the target replays", () => {
+  it("should correctly count the target breaks", () => {
```
vinceau, line 4: Can we also include a test that the target breaks array is empty if it's not a target test game replay?

vinceau (changes requested):

test/homeRun.spec.ts
cnkeats: I've included a replay from the game set to Japanese mode

Note that while the calculation in the test is correct, Japanese HRC replays currently do not play back correctly as the game is put into English mode for the replays.

vinceau (approved):

src/stats/targets.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    handleTargetBreak(allFrames, frame, this.targetBreaks);
+  }
+
+  public fetch(): TargetBreakType[] {
+    return this.targetBreaks;
+  }
+}
+
+function handleTargetBreak(frames: FramesType, frame: FrameEntryType, targetBreaks: TargetBreakType[]): boolean {
```
NikhilNarayana, line 32: does this need to return a boolean? it isn't being used anymore in the one place it gets called

NikhilNarayana (changes requested):

NikhilNarayana (approved):

PR #103, opened 2022-04-13
Removed moment package, using Date and date-fns

### LINKED ISSUE #101: Remove moment dependency
We currently only use the `moment` library for two things:
1. calling `moment().toISOString()`
2. formatting the initial filename for SLP files

We do not need to import the entirety of `moment` just to use this functionality. We can simply use `(new Date()).toISOString()` and `import { format } from 'date-fns'` to format the date keeping in mind that the [format string is 🚨**different**🚨 to `moment`](https://date-fns.org/v2.28.0/docs/format).

shrianshChari:
Fixes #101
Tested with yarn, passed all test suites and tests

### DIFF
--- a/src/utils/slpFileWriter.ts
+++ b/src/utils/slpFileWriter.ts
@@ -11,8 +11,8 @@ import { format } from "date-fns";
 /**
  * The default function to use for generating new SLP files.
  */
-function getNewFilePath(folder: string, d: Date): string {
-  return path.join(folder, `Game_${format(d, "yyyymmdd")}T${format(d, "HHmmss")}.slp`);
+function getNewFilePath(folder: string, date: Date): string {
+  return path.join(folder, `Game_${format(date, "yyyyMMdd")}T${format(date, "HHmmss")}.slp`);
 }
 
 export interface SlpFileWriterOptions extends Partial<SlpStreamSettings> {
--- a/src/utils/slpFileWriter.ts
+++ b/src/utils/slpFileWriter.ts
@@ -1,5 +1,3 @@
-import type { Moment } from "moment";
-import moment from "moment";
 import path from "path";
 import type { WritableOptions } from "stream";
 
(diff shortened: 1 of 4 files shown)


src/utils/slpFileWriter.ts
```diff
@@ -8,18 +6,20 @@ import { SlpFile } from "./slpFile";
...
 import { SlpStream, SlpStreamEvent } from "./slpStream";
 
+import { format } from "date-fns";
+
 /**
  * The default function to use for generating new SLP files.
  */
-function getNewFilePath(folder: string, m: Moment): string {
-  return path.join(folder, `Game_${m.format("YYYYMMDD")}T${m.format("HHmmss")}.slp`);
+function getNewFilePath(folder: string, d: Date): string {
```
vinceau, line 14:
```suggestion
function getNewFilePath(folder: string, date: Date): string {
```
I know you just copied it from the old function but we should probably give this a proper variable name.

src/utils/slpFileWriter.ts
```diff
@@ -8,18 +6,20 @@ import { SlpFile } from "./slpFile";
...
+  return path.join(folder, `Game_${format(d, "yyyymmdd")}T${format(d, "HHmmss")}.slp`);
```
vinceau, line 15:
```suggestion
  return path.join(folder, `Game_${format(d, "yyyyMMdd")}T${format(d, "HHmmss")}.slp`);
```
Make sure we don't confuse month and minutes.

vinceau (changes requested):
Thanks for taking this one! Just some minor changes.

src/utils/slpFileWriter.ts
shrianshChari: Yeah that probably makes sense, I'll finish refactoring on my machine

src/utils/slpFileWriter.ts
shrianshChari: Right

vinceau (approved):
LGTM! Thanks for contributing!

PR #113, opened 2022-09-08
feat: extract gecko list

JLaferri:
Extracts the gecko list from the replay, will make it easier to check if the gecko codes for two files matches


src/utils/slpReader.ts
```diff
@@ -211,12 +218,13 @@ export function iterateEvents(
...
   // Generate read buffers for each
   const commandPayloadBuffers = mapValues(slpFile.messageSizes, (size) => new Uint8Array(size + 1));
+  let splitMessageBuffer = new Uint8Array(0);
 
   const commandByteBuffer = new Uint8Array(1);
   while (readPosition < stopReadingAt) {
     readRef(ref, commandByteBuffer, 0, 1, readPosition);
-    const commandByte = commandByteBuffer[0] as number;
-    const buffer = commandPayloadBuffers[commandByte];
+    let commandByte = (commandByteBuffer[0] as number) ?? 0;
```
vinceau, line 226: What are the implications of this default commandByte `0`? Why would the command byte be undefined?

src/utils/slpReader.ts
```diff
@@ -226,14 +234,46 @@ export function iterateEvents(
...
+    const advanceAmount = buffer.length;
+
     readRef(ref, buffer, 0, buffer.length, readPosition);
+    if (commandByte === Command.SPLIT_MESSAGE) {
+      // Here we have a split message, we will collect data from them until the last
+      // message of the list is received
+      const view = new DataView(buffer.buffer);
+      const size = readUint16(view, 0x201) ?? 512;
+      const isLastMessage = readBool(view, 0x204);
+      const internalCommand = readUint8(view, 0x203) ?? 0;
```
vinceau, line 246: Over here too, what when would it be undefined to need a default value of `0`?

vinceau (approved):
LGTM but just have some questions.

src/utils/slpReader.ts
JLaferri: Under any normal circumstance it wouldn't be. It's just a fallback value to appease the type system. Actually with this the `as number` could be removed.

src/utils/slpReader.ts
JLaferri: Same as before, the `readUint8` function can return null, this ensures that the value of the variable ends up being a number.

JLaferri: Probably the more correct thing to do is to check for null and throw an error maybe? Or skip processing? But I think by setting the command to zero it will skip processing anyway. But yeah idk, I agree it could be slightly confusing.

PR #118, opened 2022-12-21
refactor: Index -> Num

NikhilNarayana:


cnkeats (approved):

src/types.ts
```diff
@@ -75,8 +75,8 @@ export interface GameStartType {
 
 interface MatchInfo {
   matchId: string | null;
-  gameIndex: number | null;
-  tiebreakerIndex: number | null;
+  gameNum: number | null;
+  tiebreakerNum: number | null;
```
vinceau, line 79:
```suggestion
  gameNumber: number | null;
  tieBreakerNumber: number | null;
```
Thoughts on using the full word instead of a the shorthand? I feel like for public interfaces it's probably better to use full words. Also any thoughts on changing `tiebreaker` to `tieBreaker`?

vinceau (approved):

PR #123, opened 2023-03-10
[PR] Correctly Calculate Winners for Timeout

JLaferri:
This fetches the final post frame updates and checks the stocks and percents on the last frame in the case of a timeout.

### DIFF
--- a/src/utils/slpReader.ts
+++ b/src/utils/slpReader.ts
@@ -1,7 +1,7 @@
 import { decode } from "@shelacek/ubjson";
 import fs from "fs";
 import iconv from "iconv-lite";
-import { mapValues } from "lodash";
+import { mapValues, isUndefined, isNull } from "lodash";
 
 import type {
   EventCallbackFunc,
@@ -12,6 +12,7 @@ import type {
   MetadataType,
   PlacementType,
   PlayerType,
+  PostFrameUpdateType,
   SelfInducedSpeedsType,
 } from "../types";
 import { Command } from "../types";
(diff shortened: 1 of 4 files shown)


src/utils/slpReader.ts
```diff
@@ -684,3 +685,49 @@ export function getGameEnd(slpFile: SlpFileType): GameEndType | null {
...
+export function extractFinalPostFrameUpdates(slpFile: SlpFileType): PostFrameUpdateType[] {
+  const { ref, rawDataPosition, rawDataLength, messageSizes } = slpFile;
+
+  // The following should exist on all replay versions
+  const postFramePayloadSize = messageSizes[Command.POST_FRAME_UPDATE];
+  const gameEndPayloadSize = messageSizes[Command.GAME_END];
+  const frameBookendPayloadSize = messageSizes[Command.FRAME_BOOKEND];
+
+  // Technically this should not be possible
+  if (isUndefined(postFramePayloadSize)) {
```
vinceau, line 698: If we don't need to distinguish between null and undefined we can use `exists` function which has already been imported.
```suggestion
  if (exists(postFramePayloadSize)) {
```

src/utils/slpReader.ts
```diff
@@ -684,3 +685,49 @@ export function getGameEnd(slpFile: SlpFileType): GameEndType | null {
...
   return gameEndMessage as GameEndType;
 }
+
+export function extractFinalPostFrameUpdates(slpFile: SlpFileType): PostFrameUpdateType[] {
+  const { ref, rawDataPosition, rawDataLength, messageSizes } = slpFile;
+
+  // The following should exist on all replay versions
+  const postFramePayloadSize = messageSizes[Command.POST_FRAME_UPDATE];
+  const gameEndPayloadSize = messageSizes[Command.GAME_END];
+  const frameBookendPayloadSize = messageSizes[Command.FRAME_BOOKEND];
```
vinceau, line 695: Frame bookend message was only introduced in version 3.0.0 onwards, what happens if we try to extract the final post frame updates without this message size?

src/utils/slpReader.ts
```diff
@@ -684,3 +685,49 @@ export function getGameEnd(slpFile: SlpFileType): GameEndType | null {
...
+    if (buffer[0] !== Command.POST_FRAME_UPDATE) {
+      break;
+    }
+
+    const postFrameMessage = parseMessage(Command.POST_FRAME_UPDATE, buffer) as PostFrameUpdateType | null;
+    if (!postFrameMessage) {
+      break;
+    }
+
+    if (isNull(frameNum)) {
```
vinceau, line 721:
```suggestion
    if (frameNum === null) {
```
nit: why not just compare with null directly here?

src/utils/slpReader.ts
```diff
@@ -684,3 +685,49 @@ export function getGameEnd(slpFile: SlpFileType): GameEndType | null {
...
+  // Technically this should not be possible
+  if (isUndefined(postFramePayloadSize)) {
+    return [];
+  }
+
+  const gameEndSize = gameEndPayloadSize ? gameEndPayloadSize + 1 : 0;
+  const postFrameSize = postFramePayloadSize + 1;
+  const frameBookendSize = frameBookendPayloadSize ? frameBookendPayloadSize + 1 : 0;
+
+  let frameNum = null;
```
vinceau, line 706:
```suggestion
  let frameNum: number | null = null;
```
We need to type this or it defaults to just an `any` type.

vinceau (changes requested):
Looking good for the most part just a few questions and comments. Were you also intending to add a test? I noticed you added an SLP file in the `placementsTest` but no accompanying test.

JLaferri:
> Looking good for the most part just a few questions and comments. Were you also intending to add a test? I noticed you added an SLP file in the `placementsTest` but no accompanying test.

Good catch, I had accidentally deleted the test. It's back now

src/utils/slpReader.ts
JLaferri: In theory `frameBookendPayloadSize` should evaluate as undefined which will later cause the `frameBookendSize` variable to evaluate to zero. In short, it should work fine I think.

src/utils/slpReader.ts
JLaferri: Why do we have our own exists function?

src/utils/slpReader.ts
JLaferri: I kind of like the readability of the first slightly better but it doesn't matter much to me. I can change it.

src/utils/slpReader.ts
vinceau: Neither `isNull`, `isNil` or `isUndefined` have been used in the codebase, so by importing and using them, we're introducing another dependency. I'd prefer it if we just used our existing `exists` function. We have our own exists function because it's convenient to handle all the null/undefined cases and it does proper type assertions too.

JLaferri: Does the dependency thing matter? It's likely an extremely lightweight function and we're already using lodash in a bunch of places?

In any case, I'll change it to `exists`.

I generally don't like writing our own functions for these kinds of things because we might not consider some edge cases that could cause strange behavior and bugs down the line; I find it better to leave it up to a mature codebase you can be pretty confident will always work right. Also it's an unnecessary function in the codebase you just have to know exists. But regardless, don't feel strongly enough about it to fight for it either.

vinceau: The `exists` function is simple enough to maintain and since we have [tree shaking enabled](https://github.com/project-slippi/slippi-js/pull/89), every additional function we import will increase the bundle size. We might not think this is a big deal but an increase in our bundle size affects and increases the size of every library/project that chooses to import our library.

vinceau (approved):
LGTM

PR #124, opened 2023-03-17
Parse instance id info

NickCondron:
This PR is intended to parse instance id data added by https://github.com/project-slippi/slippi-ssbm-asm/pull/120

Here is what a ~10 second match against a CPU looks like when you print all the post-frame/item updates and their relevant instance id info:

https://pastebin.com/5zP8MdgB

As mentioned in the other PR these fields can be used to better track hits in doubles because lastHitBy field is unreliable (https://github.com/project-slippi/slippi-js/pull/71).

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -206,6 +206,8 @@ export type PostFrameUpdateType = {
   selfInducedSpeeds: SelfInducedSpeedsType | null;
   hitlagRemaining: number | null;
   animationIndex: number | null;
+  instanceHitBy: number | null;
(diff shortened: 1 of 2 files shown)


NickCondron:
Small PR hoping to get it merged now that the instance fields are released

vinceau:
@NickCondron Do you mind adding a test and a minimal SLP (if an SLP containing the data to test doesn't already exist)?

vinceau (approved):
Actually maybe we don't need tests.

PR #125, opened 2023-03-20
Fixed jab reset detection in combos

Walnut356:
As noted by Lunar Melee, jab resets aren't picked up by the combo/conversion computers. There's also some false-detection around mangifying-glass damage.

Jab reset states added to isDamaged should take care of the former, the latter is finnicky.

An alternative solution that would fix both would be to check the hitstun (or defender hitlag) bitflag. Doing so would require replays of at least version 2.0.0, so I've held off for now in case I think of something with better compatibility.

### DIFF
--- a/src/SlippiGame.ts
+++ b/src/SlippiGame.ts
@@ -22,19 +22,19 @@ import type {
   PostFrameUpdateType,
   RollbackFrames,
 } from "./types";
-import { GameMode, GameEndMethod } from "./types";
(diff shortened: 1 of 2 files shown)


vinceau:
hey mate, thanks for this PR but your build is failing so you probably need to first run `yarn run lint --fix` to before CI will pass

vinceau (approved):

PR #135, opened 2024-03-24
Parse stage events

AlexanderHarrison:
For [this pr](https://github.com/project-slippi/slippi-ssbm-asm/pull/125) recording stage events.


vinceau (changes requested):

slp/FODPlatforms.slp
vinceau, line 1: This file would need a rename too since you've now referenced `src/FodPlatforms.slp`

vinceau (changes requested):
I'm not really familiar with the changes in the ASM repo but I would have thought maybe a single `STAGE_TRANSFORMATION_EVENT` or something that can be used for all the different stages might be better. Not sure how that would be done, but I just fear that the `Command` Enum will essentially balloon like crazy if each possible stage event has its own unique `Command`.

Not sure if @JLaferri wants to chime in.

AlexanderHarrison:
Paging @NickCondron, the implementer of stage events.

Each stage event has its own command byte, and the `Command` enum mirrors that like every other event. Although it is technically possible for the stage events to be under a single command byte, it drastically simplifies the implementation and parsing to separate them. And I don't think we are in any danger of running out of command bytes.

NickCondron:
> I'm not really familiar with the changes in the ASM repo but I would have thought maybe a single `STAGE_TRANSFORMATION_EVENT` or something that can be used for all the different stages might be better. 

Each stage has unique semantic information. You could unify them into a single STAGE_TRANSFORMATION_EVENT, but then you'd need an interior discriminant/subtype field which kinda brings you back to square one. They are unique commands in the ASM because they each have their own injection point and it makes the ASM simpler to do it that way.

The implementation in this PR looks reasonable to me at first glance. I would only add to make sure that the code handles the case where both FoD platforms are moving at the same time because you will get two events per frame in that case.

> Not sure how that would be done, but I just fear that the `Command` Enum will essentially balloon like crazy if each possible stage event has its own unique `Command`.

Personally, I don't foresee other stages/hazards being added. Yoshi's Story has Shy Guys, but they get captured in the items event and Randall is predictable based on the timer. FD and BF don't have hazards. The number of people with the ability and desire to do this for non-competitive stages is approximately zero lol.

JLaferri:
Yeah the separate commands is how it was done in the ASM and I'm cool with it. I don't really expect many (any?) more to be added and it allows for the stage events to each have the correct size, making the final SLP smaller in some cases.

vinceau (approved):
Great! Thanks for the clarifications, making those changes, and for adding tests too! Happy for this to be merged when the ASM/Dolphin changes are in :)

src/types.ts
```diff
@@ -274,6 +277,50 @@ export type GeckoCodeType = {
   contents: Uint8Array;
 };
 
+export enum FodPlatformSide {
+  RIGHT = 0,
+  LEFT = 1,
+}
+
+export type FodPlatformType = {
+  frame: number;
```
NikhilNarayana, line 286:
```suggestion
  frame: number | null;
```

all the new type fields need a similar change

NikhilNarayana (changes requested):

PR #140, opened 2024-10-09
Update getWinner() to calculate winner for legacy slps

kschniedergers:
Updated getWinner() to calculate the winner of slps before 3.13.0 that don't include the positions in the gameEnd payload. Let me know if the flow makes sense or if I should reformat the code somehow. Also this only supports singles for now.


vinceau:
I think this is not needed anymore because of the refactor completed in #165

PR #141, opened 2025-05-14
refactor: migrate yarn -> npm and update ci to node 20

NikhilNarayana:


vinceau (changes requested):
CI seems to be broken still. Maybe because we're still using `yarn` in our `.github/workflows`

vinceau (approved):
Maybe we have to merge it before CI will work again IDK.

PR #148, opened 2025-09-10
Add edge cancel stat tracking and stop counting edge cancels as failed L cancels.

mister-person:
Most people who want to improve their L cancel percentage will practice the game more. I, however, made this PR.
I took framedata.json from here: https://sourceforge.net/p/slippi-l-cancel-stats/code/ci/default/tree/framedata.json


src/stats/actions.ts
```diff
@@ -120,8 +127,8 @@ function isGrabbing(animation: State): boolean {
   return animation === State.GRAB || animation === State.DASH_GRAB;
 }
 
-function isAerialAttack(animation: State): boolean {
-  return animation >= State.AERIAL_ATTACK_START && animation <= State.AERIAL_ATTACK_END;
+function isAerialLanding(animation: State): boolean {
+  return animation >= State.AERIAL_LANDING_START && animation <= State.AERIAL_ATTACK_END;
```
JLaferri: This is a question I have about the previous iteration of this code too but why does this check for the attack range when it looks at currentAnimation below? Am I confused thinking that it should be checking prevAnimation to detect l-cancel status? I feel like this might have been a regression.

Either way at this point I think we should have both a `isAerialAttack` and `isAerialLanding` function instead of this weird function that checks both. And then we can use the one we want properly. That said I don't think we even need `isAerialAttack` at all?

mister-person: The current and previous code is checking `playerFrame.lCancelStatus` to determine if an L cancel happened, which is only set when you land with an aerial, and never during the aerial itself. Idk why the previous check included aerial animations, it made no sense. 
I just did some tests on my replay files and `lCancelStatus` also gets set in some other cases after you land, so the check still needs to be there.

JLaferri: But I mean why aren't we checking for `LANDING_START` to `LANDING_END`. Do we care about the l cancel status flag in the case where the player doesn't land? I would say no. As it is right now I feel like it will be checking the flag at very weird times. I don't even understand how it could be accurate?

It seems to me like we should only check it when previous animation is an aerial attack and current animation is landing.

mister-person: Oh, `AERIAL_ATTACK_END` should be named `AERIAL_LANDING_END`, it's just named badly bc that's what it was named before and I didn't change it. Aerials are 0x41 - 0x45, aerial landings are 0x46 - 0x4a, `AERIAL_ATTACK_START`, `AERIAL_ATTACK_END` used to be 0x41, 0x4a, so the whole range. `AERIAL_LANDING_START` to `AERIAL_ATTACK_END` is just the aerial landings. I've just renamed it.

JLaferri: Ah okay, yeah that makes more sense now. Thanks.

src/stats/actions.ts
```diff
@@ -264,20 +282,21 @@ function handleActionCompute(state: PlayerActionState, indices: PlayerIndexedTyp
     incrementCount("lCancelCount.fail", playerFrame.lCancelStatus === 2);
   }
 
+  // check for edge canceled aerial
   if (
     isAerialLanding(prevAnimation) &&
     (currentAnimation === State.FALL || currentAnimation === State.TEETER) &&
-    playerFrame.internalCharacterId
+    playerFrame.internalCharacterId !== null
```
JLaferri: Maybe you should use this instead: https://github.com/project-slippi/slippi-js/blob/5a46672b49499d1b650d4d79a6250162b1a5b981/src/utils/exists.ts

I think you probably dont want undefined either.

src/stats/actions.ts
```diff
@@ -233,13 +277,36 @@ function handleActionCompute(state: PlayerActionState, indices: PlayerIndexedTyp
...
+    // make edge cancels not count as failed L cancels
+    if (landingFrames <= lCanceledLag && state.lastLCancelStatus === 2) {
+      state.playerCounts.lCancelCount.fail -= 1;
+    }
+  }
+
   // Handles wavedash detection (and waveland)
   handleActionWavedash(state.playerCounts, state.animations);
+
+  if (playerFrame.lCancelStatus && playerFrame.lCancelStatus > 0) {
```
JLaferri: I dont think it really matters that much but I assume this first check is supposed to be `exists` as well?

PR #150, opened 2025-11-13
Refactor SlippiGame to be web-compatible by default

vinceau:
# Web-compatible slippi-js

One of the biggest complaints is that this library is not web-compatible and requires lots of polyfills due to node deps like `fs`, `events`, `stream` usage as well as external node-specific packages like `enet`, and `reconnect-core`. This refactor makes it so we have two exports:

* `@slippi/slippi-js` (subset of web-compatible things only e.g. SlippiGame, stats comp, utils etc)
* `@slippi/slippi-js/node` (full existing library with additional tools like DolphinConnection, ConsoleConnection, reading fs file paths directly into SlippiGame etc)

This should make building slippi web apps much easier.

## Problem 1 - `enet` package is node-only and requires `ws` package

When you try to make a vite/react app with the current latest `@slippi/slippi-js@^7.1.0` you get this error from trying to load the `enet` module. This module is not necessary for web applications but only when trying to interface with dolphin. There is a bug in the abandonware `enet` package where it requires the `ws` package but it is not actually marked as a dependency in the `enet` package.json. This causes build errors unless you just happen to have ws installed via a transitive dependency.

Screenshot 2025-11-17 at 5 24 56 PM

### The Fix to Problem 1

We add `ws` to our list of optional dependencies. Optional because it's only required for the `/node` exports, and `enet` is only used in the `/node` exports. This means that consumers of our package will also automatically install `ws` so if they need to use `enet` via the `DolphinConnection` they will not encounter the missing `ws` dependency issue.

## Problem 2 - Node-specific imports

If we fix the `ws` issue above and try to use the latest `@slippi/slippi-js@^7.1.0` you then now get this issue -- the core of the web-incompatibility.

Screenshot 2025-11-20 at 9 23 45 AM

Node-builtins are prolific throughout the entire project with usage of `fs`, `events`, `stream` etc. This means that the consumer is forced to polyfill everything on their side to get it to work.

### The Fix to Problem 2

* We rewrite the file reading aspect to be only within `SlpFileInputRef` and have a generic `SlpInputRef` interface which the functions inside `slpReader.ts` can read from.
* We replace usages of `EventEmitter` in stats to use our own `TypedEventEmitter` which also works both in node and the browser.
* We then also define two separate entry points - one for the browser and one for node, so that we can load the correct one. This allows vite/react apps to work out of the box - no changes needed.
* We also switch from using `dts-cli` to using rollup directly since we need more custom configuration than what `dts-cli` can offer with our two entry point build system.

## Problem 3 - `require("util")` inside `@shelacek/ubjson` breaks browser compatibility

Even after re-writing everything to be web-compatible, one of our dependencies, `@shelacek/ubjson`, includes a Node-specific fallback inside its compiled output:

```js
this.h = new("undefined" != typeof TextEncoder ? TextEncoder : require("util").TextEncoder)
```

and similarly for TextDecoder.

This fallback is harmless in Node, and honestly should be harmless in the browser too, since in the browser the globals for `TextDecoder` and `TextEncoder` exist meaning the call to `require('util')` is never made. However just having this `require('util')` breaks all browser bundlers because they see `require("util")` and assume the module needs full Node polyfills. Tools like Vite, Webpack, Rollup, and esbuild then attempt to inject or resolve util, leading to large polyfills, warnings, or outright build failures—even though browsers already provide global `TextEncoder` and `TextDecoder`.

### The Fix to Problem 3
We solve this by bundling @shelacek/ubjson only in the browser build and applying a tiny transformation during our Rollup pipeline to strip the require("util") fallback entirely.

```js
  // inside the rollup config
  replace({
    preventAssignment: false,
    delimiters: ["", ""],
    values: {
      'require("util")': "{}",
      "require('util')": "{}",
    },
  }),

```

We replace it with `{}` since we know that branch is never executed since the globals exist in all modern browsers. The Node build keeps ubjson external, so the browser-only patch does not affect Node consumers.

This ensures our browser bundle contains zero references to Node builtins so that no polyfills are required on the consumer side. This makes slippi-js fully web-compatible while still preserving full functionality in the /node export.

## End Result

We can successfully use the slippi-js package without any additional configuration.

Screenshot 2025-11-18 at 9 57 40 AM

## Achieving the `/node` export

Typically if we only define the `export` in the package.json, we can only support TypeScript 4.7+, and Webpack 5+ since they understand the export field. However, it means all older projects which use `@slippi/slippi-js` will need to update their build configuration and dependencies to use the latest package.

In order to achieve universal support, we use the **same approach as @mui/material** to achieve universal compatibility across all TypeScript versions, all bundlers, and all module resolution strategies - with **zero consumer configuration required**.

### 1. Modern: `exports` Field (TypeScript 4.7+, Webpack 5+)

For modern tools, we use the standard `package.json` exports field:

```json
{
  "exports": {
    "./node": {
      "types": "./dist/index.node.d.ts",
      "import": "./dist/node/index.esm.js",
      "require": "./dist/node/index.cjs",
      "default": "./dist/node/index.esm.js"
    }
  }
}
```

### 2. Legacy: Directory + `package.json` (All Versions)

For legacy tools, we provide a `node/` directory with its own `package.json`:

```
@slippi/slippi-js/
├── package.json          # Main package config with exports
├── dist/                 # Built files
│   ├── browser/
│   ├── node/
│   ├── index.d.ts
│   └── index.node.d.ts
└── node/                 # Legacy compatibility directory
    └── package.json      # Points to ../dist/node/
```

**node/package.json:**
```json
{
  "sideEffects": false,
  "main": "../dist/node/index.cjs",
  "module": "../dist/node/index.esm.js",
  "types": "../dist/index.node.d.ts"
}
```

## How It Works

```
┌──────────────────────────────────────────────────────┐
│  Consumer Import:                                    │
│  import { SlippiGame } from "@slippi/slippi-js/node" │
└─────────────────────┬────────────────────────────────┘
                      │
          ┌───────────┴───────────┐
          │                       │
    Modern Tool            Legacy Tool
    (TS 4.7+)             (TS 3.x-4.6)
          │                       │
          ▼                       ▼
    Uses exports            Uses directory
    (fast, optimal)         (slower, but works)
          │                       │
          └───────────┬───────────┘
                      │
                      ▼
              Same functionality,
              different paths
```

### Legacy-Only Approach
- ✅ Works everywhere
- ❌ Slower
- ❌ Less secure (no encapsulation)
- ❌ Suboptimal bundling
- ❌ Won't work with future pure ESM

### Modern-Only Approach  
- ✅ Fast
- ✅ Secure
- ✅ Optimal bundling
- ✅ Future-proof
- ❌ Breaks old tools

### Dual Approach (What We Have)
- ✅ Works everywhere (legacy fallback)
- ✅ Fast for modern tools (exports optimization)
- ✅ Secure for modern tools (encapsulation)
- ✅ Optimal bundling for modern tools
- ✅ Future-proof
- ✅ **Best of both worlds**

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -59,3 +59,6 @@ typings/
 
 # PHPStorm
 .idea/
+
+# Don't bundle source maps
+*.map


vinceau:
Since this is technically a breaking change (due to imports now being different) I would recommend that we release a minor version update before merging this PR, and then only afterwards do we make a major update release (v8.0.0). This will allow other consumers of the package to still get some non-breaking updates before they decide to update to the next version.

tsconfig.json
```diff
@@ -1,9 +1,10 @@
 {
   "include": ["src", "types"],
   "compilerOptions": {
+    "target": "ES2017",
     "module": "esnext",
     "lib": ["dom", "esnext"],
-    "importHelpers": true,
+    "importHelpers": false,
```
NikhilNarayana: is there a specific reason for disabling importHelpers?

PR #162, opened 2026-01-09
feat: Add stage alias names

vinceau:
### Description

This PR adds stage name aliases in preparation for query language parsing in the launcher. It will eventually allow us to do something like `stage:fd` instead of writing out `stage:final_destination` or something of the like.


vinceau:
Closing due to moving alias definition within the query language logic in the Launcher instead.

Issue #164, opened 2026-01-23
`getWinners` should account for stock count ties


jmlee337:
(IE. both/all players lose their last stock on the same frame)

chrissolanilla:
I'm a little puzzled on what should be done in the scenario of what should happen when there is a tie. Some rulesets seem to do it differently as far as I can tell. Could you give more guidance?

jmlee337:
https://github.com/project-slippi/slippi-js/blob/5cc4df94977f0c07457ad11fb37e5922434d59cb/src/common/utils/getWinners.ts#L5

much like the existing check for a timeout tie, `getWinners` should return `[]` if all players had 0 stocks remaining in the final post frame update (in this case, percent is ignored).

vinceau:
Fixed in #165

PR #165, opened 2026-01-25
fix: winner calculation to consider when both players have no stocks on the last frame

chrissolanilla:
References issue #164 ,

This change is mean to handle when both players have lost their stock on the last frame, we try to handle it by returning [] similar to how the code already handles timeout ties. 

I haven’t added a unit test for this yet, but I’m happy to add one if you think it would be helpful. If there’s a replay that demonstrates this scenario I can base the test on that. If not then perhaps I could add a small unit test artificially creating an object where the last frame has players with no stocks.


chrissolanilla:
@jmlee337

jmlee337:
I don't have write access here, but I would suggest that you should combine the logic for timeout ties and stock ties. You don't need to check the ending type (especially since it might be `3`), but simply need to check if either:
- all players have the same stock count and truncated percent on the last frame

or:
- all players have 0 stocks on the last frame

chrissolanilla:
> I don't have write access here, but I would suggest that you should combine the logic for timeout ties and stock ties. You don't need to check the ending type (especially since it might be `3`), but simply need to check if either:
> 
>     * all players have the same stock count and truncated percent on the last frame
> 
> 
> or:
> 
>     * all players have 0 stocks on the last frame

@jmlee337  Thanks for the feedback. I updated the logic to not rely on the gameEnd type and instead just look at the final post frame state.

Now it returns [] if all players have 0 stocks on the last frame, and for 1v1s it also handles the timeout case where both players have the same stock count and truncated percent. After that, the normal time and placements logic stays the same.

Also, do you know which person with write access I should @?

vinceau:
Thanks for your contribution. I've added tests and refactored it to allow the game end to be optional.

